PR #13, opened 2023-02-27
Added cross-fetch polyfill

rschwabco:


theluk:
Can we merge it any time soon?

egoist:
I think this will break Cloudflare Worker usage https://github.com/lquixada/cross-fetch/issues/69

rschwabco:
@egoist I'm afraid we might have to live with that for the time being. 

@theluk - I'm still testing this to ensure there are no backward-compatability issues.

junhyeokkwak:
Seems like this issue was resolved at `0.0.10` but was then reintroduced in`0.0.11`

PR #21, opened 2023-03-07
Bugfix/better error handling

rschwabco:
This PR exposes errors correctly.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,13 +5,48 @@ type PineconeClientConfiguration = {
   apiKey: string
 }
 
+Error.prepareStackTrace = () => '';
+
+async function streamToArrayBuffer(stream: ReadableStream<Uint8Array>): Promise<Uint8Array> {
+  let result = new Uint8Array(0);
+  const reader = stream.getReader();
+  while (true) { // eslint-disable-line no-constant-condition
+    const { done, value } = await reader.read();
+    if (done) {
+      break;
+    }
+
+    const newResult = new Uint8Array(result.length + value.length);
+    newResult.set(result);
+    newResult.set(value, result.length);
+    result = newResult;
+  }
+  return result;
+}
+
 async function handler(func: Function, args: any) {
   try {
     return await func(args)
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -5,13 +5,48 @@ type PineconeClientConfiguration = {
...
+
 async function handler(func: Function, args: any) {
   try {
     return await func(args)
   } catch (error) {
     // @ts-ignore
-    const message = error?.response?.data?.message || null
-    throw `PineconeClient: Error calling ${func.name.replace("bound ", "")}: ${error} ${message ? `(${message})` : ''}`
+    if (error && error.response) {
+      // @ts-ignore
```
damargulis, line 33: Can we use some sort of type guard to reduce the need for the ts-ignores here?
Like would something like

`if (error instanceof ResponseError) {...}` work?

src/index.ts
```diff
@@ -5,13 +5,48 @@ type PineconeClientConfiguration = {
...
+      const body = error.response?.body as ReadableStream
+      const buffer = body && await streamToArrayBuffer(body);
+      const text = buffer && new TextDecoder().decode(buffer);
+
+      try {
+        // Handle "RAW" call errors
+        const json = text && JSON.parse(text);
+        return Promise.reject(new Error(`${json.message}`))
```
damargulis, line 41: Is it possible for body to be null/undefined?  If so, I think its possible that json is undefined here, and this will throw another error 
`json?.message` would help

src/index.ts
```diff
@@ -5,13 +5,48 @@ type PineconeClientConfiguration = {
...
+
+      } catch (e) {
+        return Promise.reject(new Error(`PineconeClient: Error calling ${func.name.replace("bound ", "")}: ${text}`))
+      }
+    }
+    else {
+      throw new Error(`PineconeClient: Error calling ${func.name.replace("bound ", "")}: ${error}`)
```
damargulis, line 48: Why do the others reject but this one just throws? Any difference?

PR #52, opened 2023-05-04
Run tests in CI using github actions

jhamon:
## Problem

This repository has some tests, but they are not being run in CI. We need to setup this integration so we can evaluate the impact of PRs and begin to collaborate with more people.

## Solution

Implement a github actions workflow to run tests.

## Type of Change

## Test Plan

Observe things running in Github Actions.

## Notes

- There's definitely some kind of test timing issue causing these tests to flake. But some tests is better than no tests.

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   run-tests:
-    name: Run unit tests
+    name: Run integration tests
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -31,7 +31,7 @@ jobs:
       - name: Run tests
         env:
           CI: true
-          API_KEY: ${{ secrets.PINECONE_API_KEY }}
-          ENVIRONMENT: ${{ secrets.PINECONE_ENVIRONMENT }}
+          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
+          PINECONE_ENVIRONMENT: ${{ secrets.PINECONE_ENVIRONMENT }}
         run: |
           npm run test
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -13,13 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


.github/workflows/PR.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: | 
+          npm install --ignore-scripts
+      - name: Build typescript
+        run: |
+          npm run build
+      - name: Run tests
+        env:
+          CI: true
+          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
+          PINECONE_ENVIRONMENT: ${{ secrets.PINECONE_ENVIRONMENT }}
```
damargulis, line 36: Which environment do you plan to run this in?

damargulis (approved):
Nice, looks great!

.github/workflows/PR.yml
jhamon: At the moment I just have them running in my personal free tier env, but need to sort out something else soon. Any suggestion?

jhamon: We should probably discuss this somewhere less public.

rschwabco (approved):
Looks great!

tests/index.test.ts
```diff
@@ -19,27 +23,34 @@ const quantity = 10
 const metric = 'dotproduct'
 const vectors = generateVectors(dimensions, quantity)
 const vectorsWithSparseValues = generateVectors(dimensions, quantity, true)
+const client: PineconeClient = new PineconeClient()
+
+beforeEach(() => {
+  jest.setTimeout(100000)
```
micmmakarov, line 29: Is this the maximum amount of time per all tests, 100 seconds?

micmmakarov (approved):

miararoy:
LGTM, great stuff @jhamon 💪

tests/index.test.ts
jhamon: I think it's the maximum time for an individual test. Some of the tests are doing time-consuming things like creating a new index, which involves waiting a bit before it is ready. In reality, most tests do not take anywhere near that long.

Issue #44, opened 2023-04-14
[Bug] A Node.js API is used (process.nextTick) which is not supported in the Edge Runtime


trevorpfiz:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

`error - unhandledRejection: Failed to initialize Pinecone Client: PineconeClient: Error getting project name: Error: A Node.js API is used (process.nextTick) which is not supported in the Edge Runtime.`

import { PineconeClient } from '@pinecone-database/pinecone';

if (!process.env.PINECONE_ENVIRONMENT || !process.env.PINECONE_API_KEY) {
  throw new Error('Pinecone environment or api key vars missing');
}

Client:
```
async function initPinecone() {
  try {
    const pinecone = new PineconeClient();

    await pinecone.init({
      environment: process.env.PINECONE_ENVIRONMENT ?? '',
      apiKey: process.env.PINECONE_API_KEY ?? '',
    });

    return pinecone;
  } catch (error) {
    throw new Error(`Failed to initialize Pinecone Client: ${error.message}`);
  }
}

export const pinecone = await initPinecone();
```

Edge function:
```
 const filter = {
    metadataId: { $in: metadataIds },
  };

  const index = pinecone.Index(PINECONE_INDEX_NAME);
  const vectorStore = await PineconeStore.fromExistingIndex(
    new OpenAIEmbeddings({ modelName: 'text-embedding-ada-002' }),
    {
      namespace: userId,
      pineconeIndex: index,
      textKey: 'text',
      filter: filter,
    }
  );
```

### Expected Behavior

I would expect the Pinecone client to work in an edge function.

### Steps To Reproduce

1. Next.js edge function
2. attempt to use Pinecone client
3. see error

### Relevant log output

```shell
error - src/utils/pinecone.ts (18:10) @ initPinecone
error - unhandledRejection: Failed to initialize Pinecone Client: PineconeClient: Error getting project name: Error: A Node.js API is used (process.nextTick) which is not supported in the Edge Runtime.
```

### Environment

```markdown
- **OS**: Ubuntu 20.04.5 LTS with WSL2
- **Language version**: Node v18.15.0
- **Pinecone client version**: "@pinecone-database/pinecone": "^0.0.12"
```

### Additional Context

[...]

PR #53, opened 2023-05-05
Bugfix/fix edge

rschwabco:
## Problem

`process.nextTick` prevents the client from working in edge environments 

## Solution

Commented out `process.nextTick` for now. This may cause text to break.

## Type of Change

- [ *] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Test Plan

Manually testing the client.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -105,7 +105,7 @@ class PineconeClient {
     let response
     try {
       // Workaround for Jest TLSWRAP issue
-      await process.nextTick(() => { });
+      // await process.nextTick(() => { });
       response = await fetch(whoami, request);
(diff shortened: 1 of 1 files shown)


jhamon (approved):
LGTM. And tests still pass. I did some light refactoring of them yesterday before having them run in CI to make sure everything was getting cleaned up between runs, maybe that helped?

jhamon:
Seems like this will address #44

jhamon:
I force-pushed because I wanted to revert the commits where you'd bumped versions and stuff, since publishing is now handled by a github actions workflow.

PR #58, opened 2023-05-19
Added utils

rschwabco:
Adds some useful utility functions

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pinecone-database/pinecone",
-  "version": "0.1.6",
+  "version": "0.1.5",
   "main": "dist/index.js",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pinecone-database/pinecone",
-  "version": "0.1.5",
+  "version": "0.1.6",
   "main": "dist/index.js",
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    return true;
+  } catch (e) {
+    throw new Error(`Error upserting vectors into index: ${e}`);
+  }
+};
+
+
+const utils = {
```
jhamon, line 84: Did you mean to include `waitUntilIndexIsReady` here?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pinecone-database/pinecone",
-  "version": "0.1.5",
```
jhamon, line 3: Can you revert this change? CI handles version bumps now.

jhamon (approved):
Just few small things noted in comments to fix + ship.

Also, if you have a little more time, I think there's a test helper very similar to `waitUntilIndexIsReady` that you could probably replace now.

package.json
rschwabco: Sweet!!

src/utils.ts
rschwabco: No, I thought that it's really mostly useful in the context of `createIndexIfNotExist`, but we could export it as well.

PR #60, opened 2023-06-08
Add prettier formatting

jhamon:
## Problem

Formatting code by hand is tedious and inconsistent.

## Solution

Add [`prettier`](https://prettier.io/) as a devDependency:
- `.prettierignore` declares generated files we want to skip formatting
- Add a job to the PR workflow to check whether code is correctly formatted
- Add a `format` command in our npm configuration

To format code, you can add an editor plugin or run `npm run format`

`prettier` is distributed with the MIT license, so it does not introduce any licensing concerns for us. Furthermore, since it is a devDependency only, it will not be included in our built distributions of the pinecone client.

## Type of Change

## Test Plan

Describe specific steps for validating this change. See new CI job in the PR workflow passing.


damargulis (approved):

micmmakarov (approved):
We need prettier as well

PR #71, opened 2023-06-26
Add typedoc and inline documentation

jhamon:
## Problem

We need a way to create inline documentation in our source code.

## Solution

- Use [typedoc](https://typedoc.org/guides/overview/) as a devDependency. This project takes heavy inspiration from the much older jsdoc project but is compatible with typescript.
- Add initial docs about client methods landed to master
- Add `@deprecated` attribute to old client.

## Type of Change

- [x] Docs

## Test Plan

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ src/run.ts
 .vscode
 .DS_Store
 scratch
+docs/
--- a/README.future.md
+++ b/README.future.md
@@ -18,16 +18,83 @@ const client = await Pinecone.createClient({
 
 It also can find information specified in the environment variables `PINECONE_API_KEY` and `PINECONE_ENVIRONMENT`.
 
-## Data Plane
+## Control Plane
+
+### Create an index
+
+The minimum required configuration to create an index is the index name and dimension. The dimension you choose should match the output of the model used to produce your embeddings.
+
+```
+await client.createIndex({ name: 'my-index', dimension: 128 })
+```
+
+In a more expansive example, you can specify the metric, number of pods, number of replicas, and pod type.
+
+```
+await client.createIndex({
+ name: 'my-index',
+ dimension: 128,
+ metric: 'cosine',
+ pods: 1,
+ replicas: 2,
+ podType: 'p1.x1'
+})
+```
+
+By default all metadata fields are indexed when vectors are upserted with metadata, but if you want to improve performance you can specify the specific fields you want to index. This example is showing a few hypothetical metadata fields, but the values you'd use depend on what metadata you plan to store in Pinecone alongside your vectors.
+
+```
+await client.createIndex({ name: 'my-index', dimension: 128, metadataConfig: { 'indexed' : ['productName', 'productDescription'] }})
+```
+
+### Describe an index
+
+After you've created an index, you can view its configuration.
+
+```
+const config = await client.describeIndex('my-index')
+// {
+//   database: {
+//     name: 'my-index',
+//     metric: 'cosine',
+//     pods: 1,
+//     replicas: 1,
+//     shards: 1,
+//     podType: 'p1.x1'
+//   },
+//   status: { ready: false, state: 'Ready' }
+// }
+```
 
 ### List indexes
 
 ```
-client.listIndexes()
+const indexList = await client.listIndexes()
+// ['my-index']
 ```
 
-### Describe index
+### Delete an index
 
 ```
-client.describeIndex('my-index-name')
+await client.deleteIndex('my-index')
+```
+
+### Configure an index
+
+You can adjust the number of replicas or scale to a larger pod size (specified with `podType`). See [Pod types and sizes](https://docs.pinecone.io/docs/indexes#pods-pod-types-and-pod-sizes). You cannot downgrade pod size or change the base pod type.
+
+```
+await client.configureIndex('my-index', { replicas: 3, podType: 'p1.x2' })
+const config = await client.describeIndex('my-index')
+// {
+//   database: {
+//     name: 'my-index',
+//     metric: 'cosine',
+//     pods: 3,
+//     replicas: 3,
+//     shards: 1,
+//     podType: 'p1.x2'
+//   },
+//   status: { ready: false, state: 'ScalingUpPodType' }
+// }
 ```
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ src/run.ts
 .vscode
 .DS_Store
 scratch
+docs/
```
damargulis, line 9: If you wanted, it might be useful to check them in here, and then have a pr step that generates + checks for a diff (similar to : https://github.com/pinecone-io/auth0-config/blob/main/package.json#L4)

Especially if you plan to auto-sync these with the hosted docs, could be nice to have that extra check before anything goes public

README.future.md
```diff
@@ -18,16 +18,83 @@ const client = await Pinecone.createClient({
...
+ pods: 1,
+ replicas: 2,
+ podType: 'p1.x1'
+})
+```
+
+By default all metadata fields are indexed when vectors are upserted with metadata, but if you want to improve performance you can specify the specific fields you want to index. This example is showing a few hypothetical metadata fields, but the values you'd use depend on what metadata you plan to store in Pinecone alongside your vectors.
+
+```
+await client.createIndex({ name: 'my-index', dimension: 128, metadataConfig: { 'indexed' : ['productName', 'productDescription'] }})
```
damargulis, line 47: just fyi (don't think we should change anything here at this point) -- this probably won't be a thing for v4, but itll be silently ignored in the api for backwards compatability see https://github.com/pinecone-io/pinecone-db/pull/1556

src/client.ts
```diff
@@ -25,15 +25,142 @@ const ClientConfigurationSchema = Type.Object(
...
+   * @param options.dimension - The dimension of the index. Must be a positive integer. The number you choose here will depend on the model you are using. For example, if you are using a model that outputs 128-dimensional vectors, you should set the dimension to 128.
+   * @param options.metric - The metric of the index. The default metric is `'cosine'`. Supported metrics include `'cosine'`, `'dotproduct'`, and `'euclidean'`. To learn more about these options, see [Distance metrics](https://docs.pinecone.io/docs/indexes#distance-metrics)
+   * @param options.pods - The number of pods in the index. The default number of pods is 1.
+   * @param options.replicas - The number of replicas in the index. The default number of replicas is 1.
+   * @param options.podType - The type of pod in the index. This string should combine a base pod type (`s1`, `p1`, or `p2`) with a size (`x1`, `x2`, `x4`, or `x8`) into a string such as `p1.x1` or `s1.x4`. The default pod type is `p1.x1`. For more information on these, see this guide on [pod types and sizes](https://docs.pinecone.io/docs/indexes#pods-pod-types-and-pod-sizes)
```
damargulis, line 105: just curious -- when you say this works with typescript, would the docs generation error if these param tags don't make sense with the actual function?

damargulis (approved):
lgtm!

src/client.ts
jhamon: By "works with typescript" I mainly mean that at a basic level it can compile code containing typescript annotations without erroring. I started out trying to use jsdoc which is a similar tool but much older which had a big hand in defining the comment syntax used; unfortunately jsdoc immediately errors when parsing source code containing typescript annotations. You could potentially rube goldberg something together with jsdoc by running it on the compiled `tsc` output. But that sounded not great, and doesn't leverage at all the information contained in the types. So I decided to look elsewhere.

I'm still getting to know this tool. It does seem to display type information when it is available, but I'm not sure yet if it would actually error if your written comment conflicted with the type def. So far I haven't seen full-on errors, but this thing emits a lot of different warnings that I need to study more.

README.future.md
jhamon: This is good context. I think a lot of things will definitely need to be adjusted for v4. I should start a list.

.gitignore
jhamon: _Note: Sorry, this comment got super long. Basically me writing a docs one-pager strategy in a github comment 😂_ 

I'm exploring options on how to publish these, but it's looking less and less likely that these will be deployed to readme.io because the cost is so high for our use case. Something like static hosting on github pages can probably meet our needs, but need to figure out how I can build different doc versions for each versioned release. I decided not to check in this `docs` directory to main while I explore those options, but will revisit in the future if it makes sense.

It's complicated because versioned docs is closely tied to how we make releases and keep track of them. I think I need some kind of scheme where every time a release is made, we create a release branch that sticks around in case future doc changes are needed to correct documentation mistakes for previous releases. So, for example, today if I release it creates a new git tag for that version, like `v0.6.0`, but there's no long-lived release branch because so far we've only ever released from the tip of `main` and never had to deal with docs in our source. To support future maintenance of versioned docs I think I should probably also create a `release-v0.6.0` branch at release time which contains the `v0.6.0` tag but which also provides a way for version-specific docs to continue to receive updates post-release. Because important mistakes and omissions definitely happen.

If I had this kind of release branch structure, when building the static docs site, I would need a CI process that visits each `release-$VERSION` branch, builds the docs on that branch, and copies the output into a versioned subdirectory like `docs/v0.6.0`. In the end, I'd like to have a docs structure that looks something like this with all the versioned output tracked in a single branch (perhaps a branch called `docs` or `gh-pages`):

```
docs
├── index.html (intro, navigation to version-specific docs)
├── main (unreleased development)
│   ├── assets
│   ├── classes
│   ├── functions
│   ├── interfaces
│   ├── modules
│   └── types
├── v0.6.0
│   ├── assets
│   ├── classes
│   ├── functions
│   ├── interfaces
│   ├── modules
│   └── types
├── v0.7.0
│   ├── assets
│   ├── classes
│   ├── functions
│   ├── interfaces
│   ├── modules
│   └── types
└── v0.8.0
    ├── assets
    ├── classes
    ├── functions
    ├── interfaces
    ├── modules
    └── types
```

This is something that could run after every single commit to `main` or a `release-$VERSION` branch. In this type of approach we should probably check in CI that docs can compile without errors, but there's no need to check-in docs changes to main; we assume there could be changes and rebuild all the docs after every push to these branches.

damargulis: Gotcha, I hadn't really considered the need to show historical docs but definitely makes sense for versioned clients like this -- thanks for the all the info!

PR #73, opened 2023-08-02
[v1] Add upsert and fetch to new client

jhamon:
## Problem

Need to add `upsert` and `fetch` functionality to new client. Compared to the old client, these methods:

- Provide feedback when called with improper parameters
- Map server errors into a cleaner error with stacktrace
- Elevate the concept of namespaces using an optional chained index constructor method. By creating an object that has data operations scoped to that namespace, I believe you are less likely to encounter errors caused by inconsistently passing the optional namespace param.

## Solution

Similar to approach followed in other new client methods, I have wrapped the generated openapi client with additional typing and error handling.

To upsert:

```javascript

// Read config from env vars
const client = await Pinecone.createClient();

const vectors = [
  {
    id: '1',
    values: [0.1, 0.2, 0.3],
    sparseValues: {
        indices: [2],
        values: [0.1]
    },
    metadata: {}
  },
  {
    id: '2',
    values: [0.4, 0.5, 0.6],
    sparseValues: {
        indices: [2],
        values: [0.1]
    },
    metadata: {}
  }
]

// Target an index. The index dimension must match the vectors we are planning to upsert.
const index = client.index('index-name')

// Now we can upsert into the default namespace
await index.upsert(vectors)

// Now we can fetch upserted data by id
const data = await index.fetch(['1']) 

// If we wish to use non-default namespaces, we can chain a namespace call
// before upserting/fetching to create a copy of the client object scoped to perform
// all operations within that namespace.
const ns1 = index.namespace('ns1')
await ns1.upsert(vectors)

// Fetching takes place within the context of a namespace.
// This should return data because we upserted into this namespace.
await ns1.fetch(['1'])

// If we try to fetch these vectors in a different namespace,
// they will not be found.
const ns2 = index.namespace('ns2')
await ns2.fetch(['1']) // returns empty list of vector results 
```

## Known issues

There is a known issue affecting the underlying generated openapi client that does not properly build the request URL when fetching multiple ids at once. I think I have pinpointed the cause but will post a follow-up separately because it needs to be applied to both the new and old client and don't really want to do that in the same commit where I am adding a new feature.

## Type of Change

## Test Plan

Added unit tests, which should be run automatically.

For manual testing, run `npm run repl` to launch a node repl with credentials pulled from `.env`.

Manual test cases:
- ✅  Calling targeting an index that doesn't exist and trying to upsert/fetch displays a sensible error
- ✅ Calling upsert with no params errors with user feedback
- ✅ Upserting with malformed vector data errors with user feedback
- ✅ Fetching: when id does not exist returns empty results list
- ✅ Fetching: when id exists, but not in current namespace returns empty results list
- ✅ Fetching: called with no params errors with user feedback
- ✅ Fetching: called with improper type params errors with user feedback
- ❌ Fetching: when called with multiple ids (known issue)

### DIFF
--- a/src/data/__tests__/upsert.test.ts
+++ b/src/data/__tests__/upsert.test.ts
@@ -2,10 +2,12 @@ import { upsert } from '../upsert';
 import {
   PineconeBadRequestError,
   PineconeInternalServerError,
-  PineconeNotFoundError,
 } from '../../errors';
 import { VectorOperationsApi } from '../../pinecone-generated-ts-fetch';
-import type { UpsertOperationRequest, UpsertResponse } from '../../pinecone-generated-ts-fetch';
+import type {
+  UpsertOperationRequest,
+  UpsertResponse,
+} from '../../pinecone-generated-ts-fetch';
 
 describe('upsert', () => {
   test('calls the openapi upsert endpoint', async () => {
@@ -25,21 +27,22 @@ describe('upsert', () => {
       upsertRequest: {
         namespace: 'namespace',
         vectors: [{ id: '1', values: [1, 2, 3] }],
-      }
+      },
     });
   });
 
   describe('http error mapping', () => {
     test('when 500 occurs', async () => {
-      const fakeUpsert: (req: UpsertOperationRequest) => Promise<UpsertResponse> =
-        jest.fn().mockImplementation(() =>
-          Promise.reject({
-            response: {
-              status: 500,
-              text: () => 'backend error message',
-            },
-          })
-        );
+      const fakeUpsert: (
+        req: UpsertOperationRequest
+      ) => Promise<UpsertResponse> = jest.fn().mockImplementation(() =>
+        Promise.reject({
+          response: {
+            status: 500,
+            text: () => 'backend error message',
+          },
+        })
+      );
       const VOA = {
         upsert: fakeUpsert,
       } as VectorOperationsApi;
@@ -56,15 +59,16 @@ describe('upsert', () => {
     });
 
     test('when 400 occurs, displays server message', async () => {
-      const fakeUpsert: (req: UpsertOperationRequest) => Promise<UpsertResponse> =
-        jest.fn().mockImplementation(() =>
-          Promise.reject({
-            response: {
-              status: 400,
-              text: () => 'backend error message',
-            },
-          })
-        );
+      const fakeUpsert: (
+        req: UpsertOperationRequest
+      ) => Promise<UpsertResponse> = jest.fn().mockImplementation(() =>
+        Promise.reject({
+          response: {
+            status: 400,
+            text: () => 'backend error message',
+          },
+        })
+      );
       const VOA = {
         upsert: fakeUpsert,
       } as VectorOperationsApi;
--- a/src/data/__tests__/upsert.test.ts
+++ b/src/data/__tests__/upsert.test.ts
@@ -1,0 +1,85 @@
+import { upsert } from '../upsert';
+import {
+  PineconeBadRequestError,
+  PineconeInternalServerError,
+  PineconeNotFoundError,
+} from '../../errors';
+import { VectorOperationsApi } from '../../pinecone-generated-ts-fetch';
+import type { UpsertOperationRequest, UpsertResponse } from '../../pinecone-generated-ts-fetch';
+
+describe('upsert', () => {
+  test('calls the openapi upsert endpoint', async () => {
+    const fakeUpsert: (req: UpsertOperationRequest) => Promise<UpsertResponse> =
+      jest.fn();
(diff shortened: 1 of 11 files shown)


src/data/__tests__/upsert.test.ts
```diff
@@ -0,0 +1,89 @@
...
+          },
+        })
+      );
+      const VOA = {
+        upsert: fakeUpsert,
+      } as VectorOperationsApi;
+
+      jest.mock('../../pinecone-generated-ts-fetch', () => ({
+        VectorOperationsApi: VOA,
+      }));
```
iamannamai, line 78: If this is just a preference, or just some weird jest quirk that requires you do this, ignore me. No one's charging for lines of code, but if you want to save some of the hassle of the copy-paste effort, this might look something like this:

```
import { VectorOperationsApi } from '../../pinecone-generated-ts-fetch';

jest.mock('../../pinecone-generated-ts-fetch', () => ({
    ...jest.requireActual(../../pinecone-generated-ts-fetch'),
    VectorOperationsApi: jest.fn,
  }));
const mockedVectorOperationsApi = jest.mocked(VectorOperationsApi);  // exposes the mocked utilities!

describe('upsert', () => {
    it('should do the thing', () => {
        // set up your mocked implementation of upsert like you have been
        mockedVectorOperationsApi.mockReturnValue({upsert: fakeUpsert})
        
        // assert
    })
})
```

src/data/upsert.ts
```diff
@@ -0,0 +1,59 @@
+import { VectorOperationsApi } from '../pinecone-generated-ts-fetch';
+import type { ResponseError } from '../pinecone-generated-ts-fetch';
+import {
+  mapHttpStatusError,
+  PineconeConnectionError,
+  extractMessage,
+} from '../errors';
+import { builOptionConfigValidator } from '../validator';
```
iamannamai, line 8: **thought** **nit** I suppose we're too far in to worry about fixing this typo. Or am I just bad at English?

src/data/fetch.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Static, Type } from '@sinclair/typebox';
+
+const IdsArray = Type.Array(Type.String({ minLength: 1 }));
+export type IdsArray = Static<typeof IdsArray>;
+
+export const fetch = (api: VectorOperationsApi, namespace: string) => {
+  const validator = builOptionConfigValidator(IdsArray, 'fetch');
+
+  return async (ids: IdsArray): Promise<FetchResponse> => {
+    validator(ids);
```
iamannamai, line 22: **question** Maybe there are tests for just the validators that already exist outside the context of this PR, but what does the user see when the input fails validation? Is this scenario unit tested elsewhere (asking mostly so I can use it as a doc reference for what the outcome would be since I was curious about fetching without a namespace).

iamannamai (approved):
Nice work! Appreciate the consistency in all of this, makes it easier to follow. Also big fan of bespoke error types.

src/data/fetch.ts
jhamon: Good question. Here's an example failure:

```
> await client.index('jen2').upsert([{ id: 1 }])
Uncaught:
PineconeArgumentError: Argument to upsert has a problem. The property '0' must have required property 'values'. The property '0/id' must be string.
    at /Users/jhamon/workspace/pinecone-ts-client/dist/validator.js:46:15
    at /Users/jhamon/workspace/pinecone-ts-client/dist/validator.js:37:13
    at /Users/jhamon/workspace/pinecone-ts-client/dist/data/upsert.js:62:21
    at step (/Users/jhamon/workspace/pinecone-ts-client/dist/data/upsert.js:33:23)
    at Object.next (/Users/jhamon/workspace/pinecone-ts-client/dist/data/upsert.js:14:53)
    at /Users/jhamon/workspace/pinecone-ts-client/dist/data/upsert.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/jhamon/workspace/pinecone-ts-client/dist/data/upsert.js:4:12)
    at Index.upsert (/Users/jhamon/workspace/pinecone-ts-client/dist/data/upsert.js:57:40)
```

In some earlier functions added to the client library I wrote very extensive unit tests asserting the error message for each type of validation failure, but they are tedious tests to write and brittle if I change the requirements for the params. For these latest actions I skipped over those tests because it has started to feel like repetitive testing of the type checker inside the validator. Since I'm using standardized patterns now for validation and error mapping, the only thing new inside these actions is really defining the types for expected params. I've also been doing a lot of manual testing throughout development to sanity check nothing really wacky is getting shown.

src/data/upsert.ts
jhamon: I noticed this yesterday as well. I didn't want to blow up this diff by changing every file in the repo to fix. Will handle it separately sometime soon, I hope 😂

PR #74, opened 2023-08-02
[Bugfix] Fix fetch/delete with multiple ids

jhamon:
## Problem

The `fetch` and `delete` operations were not working as expected when passed multiple id values in an array. For example, ids could be fetched individually but not together.

```

// Read config from env vars
const client = await Pinecone.createClient();

const vectors = [
  {
    id: '1',
    values: [0.1, 0.2, 0.3],
    sparseValues: {
        indices: [2],
        values: [0.1]
    },
    metadata: {}
  },
  {
    id: '2',
    values: [0.4, 0.5, 0.6],
    sparseValues: {
        indices: [2],
        values: [0.1]
    },
    metadata: {}
  }
]

// Target an index. The index dimension must match the vectors we are planning to upsert.
const index = client.index('index-name')
await index.upsert(vectors)

await index.fetch(['1']) // returns data for vector 1
await index.fetch(['2']) // returns data for vector 2

// expected: returns data for both vectors
await index.fetch(['1', '2']) // actual: returns empty results 
```

The same issue is present in the legacy client, with modified sytnax.

## Solution

Since both clients are implemented around a core of generated openapi code, they shared the same root cause which is that the generated openapi client uses a faulty method of determining whether a value is an array before serializing query parameters into a string.

Previously this relied on the check `value instanceof Array` which is [known to produce confusing results](https://stackoverflow.com/questions/22289727/difference-between-using-array-isarray-and-instanceof-array) in a number of different situations. I copied the default implementation and modified this if-statement to use the much more reliable `Array.isArray` method.

Fortunately the fix is straightforward since the openapi generated client exposes a configuration parameter that allows us to substitute our own implementation for the query string generation function.

## Type of Change

## Test Plan

Re-run the above repoduction scenario and see the expected output.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -17,6 +17,7 @@ import {
 } from './control';
 import { Index } from './data';
 import { buildValidator } from './validator';
+import { queryParamsStringify } from './utils/queryParamsStringify';
 import { Static, Type } from '@sinclair/typebox';
 
 const ClientConfigurationSchema = Type.Object(
@@ -236,6 +237,7 @@ export class Client {
     const apiConfig: IndexOperationsApiConfigurationParameters = {
       basePath: controllerPath,
       apiKey,
+      queryParamsStringify,
     };
     const api = new IndexOperationsApi(new ApiConfiguration(apiConfig));
 
--- a/src/utils/queryParamsStringify.test.ts
+++ b/src/utils/queryParamsStringify.test.ts
@@ -1,0 +1,20 @@
+import { queryParamsStringify } from './queryParamsStringify';
+
+describe('queryParamsStringify', () => {
+  test('should stringify array params correctly', () => {
+    const params = {
+      ids: ['1', '2', '3'],
+    };
+    expect(queryParamsStringify(params)).toEqual('ids=1&ids=2&ids=3');
+  });
+
+  test('should stringify array params correctly when there are other params', () => {
+    const params = {
+      ids: ['1', '2', '3'],
+      other: 'param',
+    };
+    expect(queryParamsStringify(params)).toEqual(
+      'ids=1&ids=2&ids=3&other=param'
+    );
+  });
+});
--- a/src/utils/queryParamsStringify.ts
+++ b/src/utils/queryParamsStringify.ts
@@ -1,5 +1,10 @@
 import type { HTTPQuery } from '../pinecone-generated-ts-fetch';
 
+// Everything in this file is lifted from the generated openapi runtime.
+// I need to create a small modification of the generated queryParamStringify
+// function in order to fix an issue with array params.
+//
+// See https://github.com/pinecone-io/pinecone-ts-client/pull/74
 export function queryParamsStringify(
   params: HTTPQuery,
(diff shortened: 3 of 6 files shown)


src/utils/queryParamsStringify.ts
```diff
@@ -0,0 +1,50 @@
+import type { HTTPQuery } from '../pinecone-generated-ts-fetch';
+
+export function queryParamsStringify(
```
jhamon, line 3: I copied most of the code in this file from the [generated client code](https://github.com/pinecone-io/pinecone-ts-client/blob/main/src/pinecone-generated-ts-fetch/runtime.ts#L301)

src/utils/queryParamsStringify.ts
```diff
@@ -0,0 +1,50 @@
...
+    | Set<string | number | null | boolean>
+    | HTTPQuery,
+  keyPrefix: string = ''
+): string {
+  const fullKey = keyPrefix + (keyPrefix.length ? `[${key}]` : key);
+
+  // This is a one line change from the default querystring implementation. Checking
+  // with `Array.isArray` instead of `value instanceof Array` allows us to get the
+  // the correct behavior when stringifying array params.
+  if (Array.isArray(value)) {
```
jhamon, line 31: This if-condition is the only functional change.

src/utils/queryParamsStringify.ts
damargulis: You may want to leave a comment here explaining that + linking to the source

src/utils/queryParamsStringify.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  test('should stringify array params correctly when there are other params', () => {
+    const params = {
+      ids: ['1', '2', '3'],
+      other: 'param',
+    };
+    expect(queryParamsStringify(params)).toEqual(
+      'ids=1&ids=2&ids=3&other=param'
+    );
+  });
```
damargulis, line 19: Can you give an example here that wouldn't have passed the generated code?

src/client.ts
```diff
@@ -15,14 +15,16 @@ import {
...
 import { buildValidator } from './validator';
+import { queryParamsStringify } from './utils/queryParamsStringify';
 import { Static, Type } from '@sinclair/typebox';
 
 const ClientConfigurationSchema = Type.Object(
   {
     environment: Type.String({ minLength: 1 }),
     apiKey: Type.String({ minLength: 1 }),
-    projectId: Type.Optional(Type.String({ minLength: 1 })),
+    projectId: Type.String({ minLength: 1 }),
```
damargulis, line 27: Was it intentional to include these changes here? The README doesn't include projectId, so you should probably add it there at the same time

src/utils/queryParamsStringify.test.ts
jhamon: None of these pass with the generated code. I don't fully understand why, because I agree the online resources I found made it seem like an issue limited to iframes.

PR #76, opened 2023-08-10
Add user-agent header with client version to HTTP requests

jhamon:
## Problem

We need a way to understand which client versions are being used with our APIs.

## Solution

Add a `User-Agent` header to each request. The header includes the following information:
- The client version
- The client language (to easily distinguish from similarly named packages built for other languages)
- Platform information (operation system and version)
- Node version
- A flag to indicate whether the legacyExport `PineconeClient` is being used. To make the upcoming release non-breaking, we plan to continue exporting the legacy client from the module until backend API changes force us to remove it. But this flag will help us know what fraction of our users have migrated to the new client export as we navigate that process.

## Type of Change

## Test Plan

Check datadog dashboards to see new request headers coming through in integration testing.

Screenshot 2023-08-10 at 4 16 02 PM

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -17,7 +17,7 @@ import {
 } from './control';
 import { Index } from './data';
 import { buildValidator } from './validator';
-import { queryParamsStringify } from './utils/queryParamsStringify';
+import { queryParamsStringify, buildUserAgent } from './utils';
 import { Static, Type } from '@sinclair/typebox';
 
 const ClientConfigurationSchema = Type.Object(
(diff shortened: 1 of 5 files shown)


rohanshah18 (approved):
Looks good to me, thank you!

austin-denoble (approved):
Looks good, nice work. It'll be nice to have additional context for requests originating from the Node client.

Tested via repl and it looks like things are being piped through properly:

Screenshot 2023-08-11 at 12 21 35 PM

PR #77, opened 2023-08-11
[v1] Add delete vector functionality to new client

austin-denoble:
## Problem

The node client does not currently support deleting vectors via `/vectors/delete`. 

## Solution
- Add a new `delete.ts` file within `/src/data/`.
- Implement `deleteVector` to allow HTTPS calls to delete a single vector or a batch of vectors.
- Add a new `delete.test.ts` file to support unit tests. Covers validating request parameters, and error cases.
- [Bonus] Rename `/src/control/__tests__/deleteCollection.index.ts` -> `/src/control/__tests__/deleteCollection.test.ts`. The test runner may not have been picking this one up.

## Type of Change

## Test Plan

- Added a new unit test file to cover the new module. Run with `npm run test:unit`.
- Use `npm run repl` to test manually.

To test you'll want to upsert vectors and then call delete:

```javascript
npm run repl

// generate the client and legacy instances
await init()

// grab an index from client
const index = client.index('test-upsert-delete')

// upsert vector(s)
const vectors = [
  {
    id: 'test-vector-1',
    values: [1, 2, 3, 4, 5],
  },
  {
    id: 'test-vector-2',
    values: [1, 2, 3, 4, 5],
  },
];

await index.upsert(vectors);

// delete via id
await index.delete({ ids: ['test-vector-1'] })

// validate the delete
await index.fetch(['test-vector-1'])
// empty vector response: { vectors: {}, namespace: '' }

// test deleteAll
await index.delete({ deleteAll: true })

// validate the delete
await index.fetch(['test-vector-2'])
```


src/data/__tests__/delete.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+describe('deleteVector', () => {
+  test('calls the openapi delete endpoint, passing target namespace', async () => {
+    const mockDelete: (req: DeleteOperationRequest) => Promise<object> =
+      jest.fn();
+    const VOA = { _delete: mockDelete } as VectorOperationsApi;
+    const namespace = 'test-namespace';
+    const ids = ['123', '456', '789'];
+    const deleteAll = false;
+    const filter = {};
```
austin-denoble, line 17: I can probably move these inline, I pulled them out because I'm used to doing that for usability, but it may look cleaner inline.

jhamon (commented):
After testing I think this is good to merge. I think we should follow up by splitting this into `deleteOne`, `deleteMany`, and `deleteAll` just to simplify the overall usage.

jhamon (approved):

austin-denoble:
> After testing I think this is good to merge. I think we should follow up by splitting this into `deleteOne`, `deleteMany`, and `deleteAll` just to simplify the overall usage.

Sounds good, I like the idea of splitting that functionality into explicit functions rather than the primary one being a bit overloaded.

PR #79, opened 2023-08-12
[v1] Add update vector functionality to new client

austin-denoble:
## Problem
The new node client does not currently support updating vectors via `/vectors/update`.

## Solution
- Add a new `update.ts` file within `/src/data/`.
- Implement `update` to allow HTTPS calls to update a single vector.
- Add a new `update.test.ts` file to support unit tests. Covers validating request parameters, and error cases.
- [Bonus] Update `package.json` contributors to include myself.

## Type of Change
- [ X] New feature (non-breaking change which adds functionality)

## Test Plan
- Added a new unit test file to cover the new module. Run with `npm run test:unit`.
- Use `npm run repl` to test manually.

To test you'll want to call update on a specific vector and validate the results with fetch.

```javascript
npm run repl

// generate the client and legacy instances
await init()

// grab an index from client
const index = client.index('test-upsert-delete')

// upsert vector(s) or use and existing vector
const vectors = [
  {
    id: 'test-vector-1',
    values: [1, 2, 3, 4, 5],
  },
];
await index.upsert(vectors);

// update via id
await index.update({ id: 'test-vector-1', values: [5, 4, 3, 2, 1]})

// validate the update
await index.fetch(['test-vector-1'])
// updated vector response: {vectors: { test-vector-1: { id: "test-vector-1", values: [5,4,3,2,1]}}, namespace: ""}

// test updating setMetadata and sparseValues
await index.update({ id: 'test-vector-1', sparseValues: { indices: [15, 30, 25], values: [0.5, 0.5, 0.2]}, setMetadata: { genre: 'ambient' })
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
   },
   "contributors": [
     "Jen Hamon (https://github.com/jhamon)",
-    "Roie Schwaber-Cohen (https://github.com/rschwabco)"
+    "Roie Schwaber-Cohen (https://github.com/rschwabco)",
+    "Austin DeNoble (https://github.com/austin-denoble)"
   ],
   "license": "MIT",
   "scripts": {
--- a/src/data/__tests__/update.test.ts
+++ b/src/data/__tests__/update.test.ts
@@ -5,7 +5,6 @@ import {
 } from '../../errors';
 import { VectorOperationsApi } from '../../pinecone-generated-ts-fetch';
 import type { UpdateOperationRequest } from '../../pinecone-generated-ts-fetch';
-import { Vector } from '../upsert';
 
 describe('update', () => {
   test('calls the openapi update endpoint, passing target namespace', async () => {
--- a/src/data/__tests__/update.test.ts
+++ b/src/data/__tests__/update.test.ts
@@ -1,0 +1,97 @@
+import { update } from '../update';
+import {
+  PineconeBadRequestError,
+  PineconeInternalServerError,
+} from '../../errors';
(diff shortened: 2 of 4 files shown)


jhamon:
I tested on local and this seems good to merge. Testing covered:
- default and non-default namespace
- missing required id param
- passing incorrect data types
- using different combinations of metadata, values, sparseValues

Some UX concerns I had (that I can address in follow-ups):
- `setMetadata` name is unintuitive. I tried updating with `metadata` and it didn't work which briefly made me think it wasn't working until I looked into the code to see what I should pass. I think we should change this to accept `metadata` instead of or in addition to `setMetadata` (which is a weirdo name coming from the REST API).
- I was also surprised it didn't error when I tried updating an ID that doesn't exist. But I assume that's down to the API behavior and not something we can control within the client itself.

jhamon (approved):

austin-denoble:
> Some UX concerns I had (that I can address in follow-ups):
> 
> * `setMetadata` name is unintuitive. I tried updating with `metadata` and it didn't work which briefly made me think it wasn't working until I looked into the code to see what I should pass. I think we should change this to accept `metadata` instead of or in addition to `setMetadata` (which is a weirdo name coming from the REST API).
> * I was also surprised it didn't error when I tried updating an ID that doesn't exist. But I assume that's down to the API behavior and not something we can control within the client itself.

Definitely sound like good suggestions for improving the UX here. I liked your offline suggestion around returning a 404 in the case of the vector not existing, but we'd probably need tweaks to the API to support it.

PR #83, opened 2023-08-21
use metadata: instead of setMetadata: for index.update param;

tdonia:
## Problem

index.update() used the unintuitive `setMetadata` parameter instead of `metadata` because the underlying API and generated files use `setMetadata`; originated in #79 

## Solution

This uses `metadata` instead of `setMetadata` in the ts interface, then modifies options to use `setMetadata` for the actual request to avoid changing the API.

## Type of Change

## Test Plan
```
await index.update({id: "1", metadata: {foo: "bar"}}) //should work
await index.update({id: "1", setMetadata: {foo: "bar"}}) //shouldn't work
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ const updateRequest: UpdateRequest = {
     indices: [15, 30, 11],
     values: [0.1, 0.2, 0.3],
   }, // optional sparse values
-  setMetadata: metadata, // the new metadata
+  metadata: metadata, // the new metadata
   namespace,
 };
 await index.update({ updateRequest });
--- a/src/data/__tests__/update.test.ts
+++ b/src/data/__tests__/update.test.ts
@@ -37,7 +37,7 @@ describe('update', () => {
           indices: [15, 30, 25],
           values: [0.5, 0.5, 0.2],
         },
-        metadata: { genre: 'ambient' },
+        setMetadata: { genre: 'ambient' },
       },
     });
   });
--- a/src/data/__tests__/update.test.ts
+++ b/src/data/__tests__/update.test.ts
@@ -24,7 +24,7 @@ describe('update', () => {
         indices: [15, 30, 25],
         values: [0.5, 0.5, 0.2],
       },
-      setMetadata: { genre: 'ambient' },
+      metadata: { genre: 'ambient' },
     });
 
     expect(returned).toBe(void 0);
@@ -37,7 +37,7 @@ describe('update', () => {
           indices: [15, 30, 25],
           values: [0.5, 0.5, 0.2],
         },
-        setMetadata: { genre: 'ambient' },
+        metadata: { genre: 'ambient' },
       },
     });
   });
--- a/src/data/update.ts
+++ b/src/data/update.ts
@@ -13,7 +13,6 @@ const UpdateVectorOptionsSchema = Type.Object({
   values: Type.Optional(Type.Array(Type.Number())),
   sparseValues: Type.Optional(SparseValues),
   metadata: Type.Optional(Type.Object({}, { additionalProperties: true })),
-  setMetadata: Type.Optional(Type.Object({}, { additionalProperties: true })),
 });
 
 export type UpdateVectorOptions = Static<typeof UpdateVectorOptionsSchema>;
@@ -26,11 +25,16 @@ export const update = (api: VectorOperationsApi, namespace: string) => {
 
   return async (options: UpdateVectorOptions): Promise<void> => {
     validator(options);
-    options["setMetadata"] = options["metadata"];
-    delete(options["metadata"]);
+
+    const requestOptions = {
+      id: options['id'],
+      values: options['values'],
+      sparseValues: options['sparseValues'],
+      setMetadata: options['metadata'],
+    };
 
     try {
-      await api.update({ updateRequest: { ...options, namespace } });
+      await api.update({ updateRequest: { ...requestOptions, namespace } });
       return;
     } catch (e) {
       const err = await handleDataError(e);
--- a/src/data/update.ts
+++ b/src/data/update.ts
@@ -12,6 +12,7 @@ const UpdateVectorOptionsSchema = Type.Object({
   id: Type.String({ minLength: 1 }),
   values: Type.Optional(Type.Array(Type.Number())),
   sparseValues: Type.Optional(SparseValues),
+  metadata: Type.Optional(Type.Object({}, { additionalProperties: true })),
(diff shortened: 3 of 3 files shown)


src/data/update.ts
```diff
@@ -25,6 +26,8 @@ export const update = (api: VectorOperationsApi, namespace: string) => {
 
   return async (options: UpdateVectorOptions): Promise<void> => {
     validator(options);
+    options["setMetadata"] = options["metadata"];
```
jhamon, line 29: The way you added metadata to the type definition above implies that either `metadata` or `setMetadata` may be passed, but here in the implementation you are always overriding the value of setMetadata with the value of metadata (even if metadata was not passed). So there's a scenario where a person has called update passing a value in `setMetadata` that will then be overwritten as `undefined` before the request is made. E.g. something along the lines `client.index('my_index').update({id: '1', values: [0.1, 0.2, 0.3], setMetadata: { indexed: [ 'propertyName' ]}} )`. Even if we won't be showing the usage of setMetadata in the future, people who are migrating old code to use the new client could plausibly pass a value under the setMetadata key.

I would remove `setMetadata` from the type definition above to communicate our intention that `metadata` is what should be passed. And then I would wrap this assignment line `options["setMetadata"] = options["metadata"];` in an if statement that checks whether a value of setMetadata has been passed before overriding it. This is necessary because our typing is open so that if any new properties are added to the API in the future then they can immediately be used without updating the client.

src/data/__tests__/update.test.ts
```diff
@@ -37,7 +37,7 @@ describe('update', () => {
           indices: [15, 30, 25],
           values: [0.5, 0.5, 0.2],
         },
-        setMetadata: { genre: 'ambient' },
+        metadata: { genre: 'ambient' },
```
jhamon, line 40: I think this part of the test should not change, since we're still sending the same request to the backend API.

README.md
```diff
@@ -244,7 +244,7 @@ const updateRequest: UpdateRequest = {
     indices: [15, 30, 11],
     values: [0.1, 0.2, 0.3],
   }, // optional sparse values
-  setMetadata: metadata, // the new metadata
+  metadata: metadata, // the new metadata
```
jhamon, line 247: I'm planning to rewrite this entire README ahead of the release. i don't want people to find instructions for unpublished code in the README. But I appreciate the instinct to update docs as you go.

jhamon (approved):
You can easily resolve the lint errors in CI by running `npm run format` and committing the formatting changes.

I left some minor feedback on ways to adjust the implementation. Let me know if you'd like another review, but otherwise I trust you to fix and merge.

tdonia:
thank you, @jhamon!  per your [comment](https://github.com/pinecone-io/pinecone-ts-client/pull/83#discussion_r1300735305), i removed `setMetadata` from `UpdateVectorOptionsSchema` so it will no longer validate. does that approach make sense or did i miss something?

src/data/update.ts
tdonia: > This is necessary because our typing is open so that if any new properties are added to the API in the future then they can immediately be used without updating the client.

does this mean that we want to preserve the ability for `client.index('my_index').update({id: '1', values: [0.1, 0.2, 0.3], somethingTotallyNew: "foo" } )` to work with no client updates?  doesn't that fail because `somethingTotallyNew` won't validate?

src/data/update.ts
```diff
@@ -26,8 +26,15 @@ export const update = (api: VectorOperationsApi, namespace: string) => {
   return async (options: UpdateVectorOptions): Promise<void> => {
     validator(options);
 
+    const requestOptions = {
+      id: options['id'],
+      values: options['values'],
+      sparseValues: options['sparseValues'],
+      setMetadata: options['metadata'],
+    };
```
austin-denoble, line 34: **praise:** I like this for re-mapping things. 👍

austin-denoble (approved):
LGTM, thanks for getting this updated to be a bit more clear!

src/data/update.ts
jhamon: Currently calling this method with `somethingTotallynew: "foo"` would pass validation because our `UpdateVectorOptionsSchema` type definition does not include `{ additionalProperties: false }`. So it's a loose typing providing enforcement of the presence of required properties and the types of required and optional properties. Essentially, if you do not pass something we expect (e.g. you leave out the `id` property, or pass an `id` with incorrect type, the validator will reject it. But if you pass something new or unknown (or, unfortunately, if you misspell something optional), you won't get an error. 

The flexibility this affords to update properties that may not exist when this client is released is nice, but it comes at the cost of good feedback in some situations. For example, if you accidentally misspell the name of an optional property such as `metadata` as `meatdata`, you will not get feedback from the validator.

It's a tough call between the benefits of stricter typing vs the flexibility to tolerate certain kinds of changes in the API. For now we've stayed on the side of flexibility, but if you think we should take a more rigid stance then we can talk about it more.

tdonia: that makes sense, and also means my [revised implementation](https://github.com/pinecone-io/pinecone-ts-client/pull/83/files/c72c5f58eb16335f98b4ef67ca797fc8daf8eb89#diff-6910af72b7db376da51a11a1638b35d8910dcb07662cd625fb17293acf34bf65L28) is going to break that principle because it assumes we know all of the request params.  i'll take another pass at it!

src/data/update.ts
tdonia: thanks! me too! but i'm going to revise again so as to not break the principle Jen mentions [here](https://github.com/pinecone-io/pinecone-ts-client/pull/83#discussion_r1303365302)

jhamon: This reads nicely, but I think it means we lose the flexibility of being able to update any properties that get added in the future.

austin-denoble: Got it, catching up on the follow-up comments.

src/data/update.ts
tdonia: following up based on an offline conversation -- that principle doesn't hold because our API requests have typed parameters, so the [implementation here](https://github.com/pinecone-io/pinecone-ts-client/pull/83/files/c72c5f58eb16335f98b4ef67ca797fc8daf8eb89#diff-6910af72b7db376da51a11a1638b35d8910dcb07662cd625fb17293acf34bf65L28) will work.

src/data/update.ts
tdonia: one more follow-up -- this _is_ okay for now because our generated API code doesn't allow the forward flexibility we thought it did, so we'll keep this implementation.

PR #86, opened 2023-08-24
[v1] Split delete functionality into deleteOne, deleteMany, and deleteAll

austin-denoble:
## Problem
Currently, the `delete` method for handling vectors is a bit overloaded. The API accepts a variety of parameters for allowing flexibility in deleting vectors allowing you to pass a list of `ids`, a boolean for `deleteAll`, and a `filter` object for deleting vectors via their metadata:

The different fields for controlling deletions have a priority on the backend https://github.com/pinecone-io/pinecone-db/blob/8da299373c086af48cc02cda5ace33703490cb46/pinecone-db/src/bin/replica.rs#L236. 

The user may also receive errors in response if they compose the delete with multiple fields such as `ids` with `deleteAll: true`: https://github.com/pinecone-io/pinecone-db/blob/8da299373c086af48cc02cda5ace33703490cb46/pinecone-db/src/server/validator.rs#L100

We want to lead users into a pit of success by splitting out the delete functionality into discrete actions which are a bit more clear.

## Solution
- Create new files under `/data/` to explicitly handle dataplane deletion operations: `deleteOne.ts`, `deleteMany.ts`, `deleteAll.ts`. Create explicit typebox schemas for each function to better control what the user is able to pass in.
- Add new unit tests for each specific file to verify parameters being sent with namespace.
- Remove the existing `delete.ts` file and functionality from the API.

## Type of Change

## Test Plan
New unit tests for covering each delete function.

### Manual Testing

`deleteAll`
```
npm run repl
await init()

// upsert vectors for testing 
await client.index('test-index').upsert([{ id: 'test-vector-1', values: [1, 2, 3] }, { id: 'test-vector-2', values: [3, 2, 1] }])

// delete all and validate
await client.index('test-index').deleteAll()

await client.index('test-index').fetch(['test-vector-1', 'test-vector2'])
{ vectors: {}, namespace: '' }
```

`deleteMany`
```
npm run repl
await init()

// upsert vectors for testing including some with different metadata structures
const vectors = [
  { id: 'test-vector-1', values: [1, 2, 3] },
  { id: 'test-vector-2', values: [3, 2, 1], metadata: { genre: 'rock' } },
  { id: 'test-vector-3', values: [2, 4, 1], metadata: { genre: 'ambient' } },
  { id: 'test-vector-4', values: [5, 1, 3], metadata: { genre: 'pop' } },
];
await client.index('test-index').upsert(vectors)

// test deleting via filter and validate
await client.index('test-index').deleteMany({ filter: { genre: 'rock' })
await client.index('test-index').fetch(['test-vector-2'])
{ vectors: {}, namespace: '' }

// test deleting via ids list and validate
await client.index('test-index').deleteMany({ ids: ['test-vector-3', 'test-vector-4'] })
await client.index('test-index').fetch(['test-vector-3', 'test-vector-4'])
{ vectors: {}, namespace: '' }

// test passing both filter and ids to make sure AJV errors out as expected
await client.index('test-index').deleteMany({ ids: ['test-vector-3', 'test-vector-4'], filter: { genre: 'ambient' } })
Uncaught:
PineconeArgumentError: The argument to deleteMany accepts multiple types. Either 1) 2)
    at /Users/austin/workspace/pinecone-ts-client/dist/validator.js:190:19
    at /Users/austin/workspace/pinecone-ts-client/dist/data/deleteMany.js:75:21
    at step (/Users/austin/workspace/pinecone-ts-client/dist/data/deleteMany.js:44:23)
    at Object.next (/Users/austin/workspace/pinecone-ts-client/dist/data/deleteMany.js:25:53)
    at /Users/austin/workspace/pinecone-ts-client/dist/data/deleteMany.js:19:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/austin/workspace/pinecone-ts-client/dist/data/deleteMany.js:15:12)
    at Index.deleteMany (/Users/austin/workspace/pinecone-ts-client/dist/data/deleteMany.js:70:40)
    at REPL62:1:45
```
**Note:** The above error will be fixed in a follow-up @jhamon is working on, there should be additional info about which types are acceptable.

`deleteOne`
```
npm run repl
await init()

// upsert vectors for testing
await client.index('test-index').upsert([{ id: '123', values: [3, 2, 1]}, { id: '321', values: [1, 2, 3]} ])

// delete a single vector and validate
await client.index('test-index').deleteOne('123')
await client.fetch('123')
{ vectors: {}, namespace: '' }
```


jhamon (approved):
I did a lot of hands on testing, including scoping down to a particular namespace, and it seems to work as I expect.

We need to export all of our typescript types for each methods arguments from the module for people to be able to use them in their own apps. I lumped in this change for other actions in my query PR (which I'm going to merge today 🤞 ) but didn't make that change for these new actions. Have a look at `src/data/index.ts` and `src/index.ts` in [my PR](https://github.com/pinecone-io/pinecone-ts-client/pull/85/files#diff-46a6fe9a2a5e2561f258a496b2df29d3c9306ae1fde832a0f99ac55f8bb8eedfR13). We should do a final audit before the release to make sure we didn't forget any.

**Idea**: What do you think about having deleteMany also accept an array of ids? Is that too much flexibility? E.g. `await client.index('foo').namespace('bar').deleteMany([ 'test-id-1', 'test-id-2' ])`

austin-denoble:
> Idea: What do you think about having deleteMany also accept an array of ids? Is that too much flexibility? E.g. await client.index('foo').namespace('bar').deleteMany([ 'test-id-1', 'test-id-2' ])

I like that, the question for me is then do you still need to pass an object with a nested `filter: { ... }` object, or could we also allow them to just pass a top-level object representing `filter`:
`await client.index('foo').namespace('bar').deleteMany({ genre: 'rock', year: 1988 })`

I'm kind of worried this is maybe more confusing than having them as fields on an object, but maybe that is just as confusing.

austin-denoble:
**Summary**

After talking with @jhamon offline we've decided to move forward with altering `deleteMany()` to support passing different arguments rather than an object with specific fields. Now, a user can simply pass an array of vector ids, or a `filter` object to support deleting many vectors. This feels like it aligns nicely with the current API offering `deleteAll()` (with no arguments), or `deleteOne('vector-id')` which accepts a string for deleting a single vector.

I've also added additional validation rules to the `filter` object that can be passed. Pinecone accepts an empty filter object (`filter: {}`) for delete calls, and will proceed to delete all vectors if this is provided. Given we offer an explicit way to do this via `deleteAll`, we want to prevent any possible footguns for users working with the deletion API. Leveraging `minProperties: 1` for the `filter` argument allows us to guide users around working with deletes.

src/control/index.ts
```diff
@@ -1,19 +1,19 @@
 // Index Operations
```
austin-denoble, line 1: **Note:** Alphabetizing our exports here.

src/data/index.ts
```diff
@@ -8,20 +8,22 @@ import { fetch } from './fetch';
 import { update } from './update';
```
austin-denoble, line 8: **Note:** Alphabetizing our exports / function hookups.

src/index.ts
```diff
@@ -4,32 +4,33 @@ export { Client, type ClientConfiguration } from './client';
 export * as Errors from './errors';
 
 // Type exports
```
austin-denoble, line 6: **Note:** Alphabetizing our exports here.

PR #87, opened 2023-08-24
Increase strictness of TypeBox Object schemas

austin-denoble:
## Problem
When creating schemas for runtime validation with typebox involving `Type.Object()`, we have not been providing the `additionalProperties: false` parameter via `ObjectOptions`. This would restrict users' ability to provide extraneous parameters at runtime when calling functions that receive objects, etc.

Initially, we had thought this flexibility would be beneficial in that users could leverage the client using fields that may have been added since the last update. However, OpenAPI prevents this behavior anyway under the hood, so we'd like to explicitly disallow the ability for users to pass additional values.

Context: https://pinecone-io.slack.com/archives/C04DQS0RG84/p1692816422329919 

## Solution

For all instances where we're leveraging `Type.Object({ ... })` with an explicit list of params, we want to make sure to provide the `additionalProperties` option explicitly: `Type.Object({ ... }, { additionalProperties: false })`. This will explicitly prevent users from passing extra params, and should alert them to this via our validation pipeline.

**Note:** I was curious what the behavior is when omitting the `additionalProperties` value entirely. My assumption was it would default to false. I dug around in the typebox code a bit to get a better sense of things:

```javascript
    function* Object(schema, references, value) {
        yield IsObjectCheck(value);
        if (IsNumber(schema.minProperties))
            yield `Object.getOwnPropertyNames(${value}).length >= ${schema.minProperties}`;
        if (IsNumber(schema.maxProperties))
            yield `Object.getOwnPropertyNames(${value}).length <= ${schema.maxProperties}`;
        const knownKeys = globalThis.Object.getOwnPropertyNames(schema.properties);
        for (const knownKey of knownKeys) {
            const memberExpression = MemberExpression.Encode(value, knownKey);
            const property = schema.properties[knownKey];
            if (schema.required && schema.required.includes(knownKey)) {
                yield* Visit(property, references, memberExpression);
                if (Types.ExtendsUndefined.Check(property) || IsAnyOrUnknown(property))
                    yield `('${knownKey}' in ${value})`;
            }
            else {
                const expression = CreateExpression(property, references, memberExpression);
                yield IsExactOptionalProperty(value, knownKey, expression);
            }
        }
        if (schema.additionalProperties === false) {
            if (schema.required && schema.required.length === knownKeys.length) {
                yield `Object.getOwnPropertyNames(${value}).length === ${knownKeys.length}`;
            }
            else {
                const keys = `[${knownKeys.map((key) => `'${key}'`).join(', ')}]`;
                yield `Object.getOwnPropertyNames(${value}).every(key => ${keys}.includes(key))`;
            }
        }
        if (typeof schema.additionalProperties === 'object') {
            const expression = CreateExpression(schema.additionalProperties, references, `${value}[key]`);
            const keys = `[${knownKeys.map((key) => `'${key}'`).join(', ')}]`;
            yield `(Object.getOwnPropertyNames(${value}).every(key => ${keys}.includes(key) || ${expression}))`;
        }
    }
```
(the above can be found in `/@sinclair/typebox/compiler/compiler.js` if you're curious)

## Type of Change

## Test Plan
Testing should involve attempting to pass objects with extra parameters to the specific functions that were affected and making sure we receive the proper validation error from AJV:
- `configureIndex`
- `createCollection`
- `createIndex`
- `update`
- `upsert`

### upsert example
```
npm run repl
await init()

> await client.index('test-create-index').upsert([{ id: 'testing', foo: 'bar', values: [1, 2, 3]}])
Uncaught:
PineconeArgumentError: The argument to upsert had validation errors: item at index 0 of the array must NOT have additional properties.
    at /Users/austin/workspace/pinecone-ts-client/dist/validator.js:191:19
    at /Users/austin/workspace/pinecone-ts-client/dist/data/upsert.js:62:21
    at step (/Users/austin/workspace/pinecone-ts-client/dist/data/upsert.js:33:23)
    at Object.next (/Users/austin/workspace/pinecone-ts-client/dist/data/upsert.js:14:53)
    at /Users/austin/workspace/pinecone-ts-client/dist/data/upsert.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/austin/workspace/pinecone-ts-client/dist/data/upsert.js:4:12)
    at Index.upsert (/Users/austin/workspace/pinecone-ts-client/dist/data/upsert.js:57:40)
    at REPL14:1:73
```
**NOTE:** It seems like there may be some issues regarding typebox union types that involve objects with restricted properties. This came up when running the `configureIndex.validation.test.ts` file after adjusting object types:

Screenshot 2023-08-24 at 4 24 00 PM

The error message ends up looking a bit confusing here. I believe this may also be addressed in a `validator.ts` fix @jhamon is currently working on. For now, I've updated the expected string in the unit test, but am open to feedback.

### DIFF
--- a/src/control/__tests__/configureIndex.validation.test.ts
+++ b/src/control/__tests__/configureIndex.validation.test.ts
@@ -72,7 +72,7 @@ describe('configureIndex argument validations', () => {
 
       expect(toThrow).rejects.toThrowError(PineconeArgumentError);
       expect(toThrow).rejects.toThrowError(
-        "The second argument to configureIndex accepts multiple types. Either 1) must have required property: replicas. 2) had type errors: property 'podType' must be string."
+        "The second argument to configureIndex accepts multiple types. Either 1) must have required property: replicas. There were also validation errors: argument must NOT have additional properties. 2) had type errors: property 'podType' must be string."
       );
     });
 
--- a/src/control/configureIndex.ts
+++ b/src/control/configureIndex.ts
@@ -11,12 +11,18 @@ const nonemptyString = Type.String({ minLength: 1 });
 const positiveInteger = Type.Integer({ minimum: 1 });
 
 const ConfigureIndexOptionsSchema = Type.Union([
-  Type.Object({
-    replicas: positiveInteger,
-  }),
-  Type.Object({
-    podType: nonemptyString,
-  }),
+  Type.Object(
+    {
+      replicas: positiveInteger,
+    },
+    { additionalProperties: false }
+  ),
+  Type.Object(
+    {
+      podType: nonemptyString,
+    },
+    { additionalProperties: false }
+  ),
 ]);
 
 export type ConfigureIndexOptions = Static<typeof ConfigureIndexOptionsSchema>;
--- a/src/control/createCollection.ts
+++ b/src/control/createCollection.ts
@@ -8,10 +8,13 @@ import { Static, Type } from '@sinclair/typebox';
 
 const nonemptyString = Type.String({ minLength: 1 });
 
-const CreateCollectionOptionsSchema = Type.Object({
-  name: nonemptyString,
-  source: nonemptyString,
-});
+const CreateCollectionOptionsSchema = Type.Object(
+  {
+    name: nonemptyString,
+    source: nonemptyString,
+  },
+  { additionalProperties: false }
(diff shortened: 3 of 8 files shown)


jhamon (approved):
LGTM 👍

PR #89, opened 2023-08-30
[v1] Allow createIndex to support waitUntilReady for async polling

austin-denoble:
## Problem
Creating a new index is an asynchronous operation on the backend. When calling `client.createIndex({ ... })` the operation to trigger creating the index occurs nearly immediately, but the index will not actually be ready for some amount of time after the creation call returns (somewhere between 10-13 seconds average currently).

v0 offered a utility function `waitUntilIndexIsReady` which allowed an async call leveraging polling `describeIndex()`.

## Solution
- Add a new `waitUntilReady` option to `createIndex()` allowing the user to opt for returning an async function that polls `describeIndex()` until the index is ready.
- Add a new `waitUntilIndexIsReady()` function in `/createIndex.ts` to handle polling `describeIndex()` if desired. Originally I was going to put the function in `/utils` but it felt specific enough to creating the index that it made more sense living here for now.
- Update `createIndex.test.ts`
   - Added some error case tests for the basic functionality.
   - Added tests to run through the asynchronous logic of `waitUntilIndexIsReady()`.
   - Upgraded `@types/jest` as there was some functionality in Jest 29.5.0 that wasn't represented. Did a lot of troubleshooting around jest and mocking timers while working on tests for this, they had [exposed a bunch of new functions recently for working with fake timers](https://github.com/jestjs/jest/pull/13981) and autocomplete lead me astray. 😭 

## Type of Change

## Test Plan
Validate that passing `waitUntilReady: true` to `createIndex()` await the creation of the index on the backend.

```
$ npm run repl

$ await init()

// Wait until index is ready
$ await client.createIndex({ name: 'test-create-1', dimension: 5, waitUntilReady: true })
'Index is ready after 13 seconds'

// Test the basic scenario of returning as soon as the create call completes
$ await client.createIndex({ name: 'test-create-2', dimension: 5 })
undefined
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "devDependencies": {
     "@jest/globals": "^29.3.1",
-    "@types/jest": "^29.2.5",
+    "@types/jest": "^29.5.0",
     "@types/node": "^18.11.17",
     "@typescript-eslint/eslint-plugin": "^5.59.11",
     "@typescript-eslint/parser": "^5.59.11",
--- a/src/control/createIndex.ts
+++ b/src/control/createIndex.ts
@@ -66,6 +66,6 @@ const waitUntilIndexIsReady = async (
       return `Index is ready after ${seconds} seconds`;
     }
   } catch (e) {
-    throw new Error('Error creating index');
+    throw new Error(`Error creating index: ${e}`);
   }
 };
--- a/src/control/createIndex.ts
+++ b/src/control/createIndex.ts
@@ -1,5 +1,7 @@
 import { IndexOperationsApi } from '../pinecone-generated-ts-fetch';
 import { buildConfigValidator } from '../validator';
+import { debugLog } from '../utils';
+import { handleApiError } from '../errors';
 import { handleIndexRequestError } from './utils';
 import { Static, Type } from '@sinclair/typebox';
 
@@ -63,9 +65,15 @@ const waitUntilIndexIsReady = async (
       await new Promise((r) => setTimeout(r, 1000));
       return await waitUntilIndexIsReady(api, indexName, seconds + 1);
     } else {
-      return `Index is ready after ${seconds} seconds`;
+      debugLog(`Index ${indexName} is ready after ${seconds}`);
+      return;
     }
   } catch (e) {
-    throw new Error(`Error creating index: ${e}`);
+    const err = await handleApiError(
+      e,
+      async (_, rawMessageText) =>
+        `Error creating index ${indexName}: ${rawMessageText}`
+    );
+    throw err;
   }
 };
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
   },
   "devDependencies": {
     "@jest/globals": "^29.3.1",
-    "@types/jest": "^29.2.5",
+    "@types/jest": "^29.5.0",
```
austin-denoble, line 30: See my comment in the PR summary. There were some Jest internals that weren't being properly exposed, I think it's somewhat safe to update our typing package (to be honest I'm not sure if this package version is specifically aligned with Jest).

src/control/createIndex.ts
```diff
@@ -54,3 +57,21 @@ export const createIndex = (api: IndexOperationsApi) => {
...
+  indexName: string,
+  seconds: number = 0
+) => {
+  try {
+    const indexDescription = await api.describeIndex({ indexName });
+    if (!indexDescription.status?.ready) {
+      await new Promise((r) => setTimeout(r, 1000));
+      return await waitUntilIndexIsReady(api, indexName, seconds + 1);
+    } else {
+      return `Index is ready after ${seconds} seconds`;
```
austin-denoble, line 72: **Note:** I'm a little unsure of how we should return out of here. Originally there were `console.log()` calls, but I wasn't sure if that was the best option either as a lot of our current calls return either undefined or the actual response entity. Definitely open to taking different approaches to both the successful return here and how we're throwing a possible error below.

jhamon: I think this should just `return;` undefined so that `createIndex` will have a consistent return type any time it is used (with or without the wait).

Debug output could be useful, though, definitely. We don't have a proper logging solution in place. For now you could just hack something with console log and an environment variable.

I might add a function like 

```
const debugLog = (str) {
      if (process && process.env && process.env.PINECONE_DEBUG) {
                 console.log(`Index ${indexName} is ready after ${seconds} seconds`;
              }
      }
}
```

It is important to check the `process` object is present before accessing it. Because people could use this in exotic contexts where that's not available.

Then I'd do something along the lines of
```
try {
    const indexDescription = await api.describeIndex({ indexName });
    if (!indexDescription.status?.ready) {
      await new Promise((r) => setTimeout(r, 1000));
      return await waitUntilIndexIsReady(api, indexName, seconds + 1);
    } else {
       debugLog(`Index ${indexName} is ready after ${seconds} seconds`)
       return;
    }
}
```

jhamon (approved):
I suggested a change, but I would consider it non-blocking if you want to merge and handle as a follow-up.

src/control/createIndex.ts
austin-denoble: Good call, I went ahead and added the `debugLog()` function to `src/utils/` with a small test file to validate the flag. If we want to expand this or make changes going forward I figured it'd be nice to situate it there for now.

PR #90, opened 2023-08-30
[v1] Allow upsert to support chunkSize for chunked vector upserts

austin-denoble:
implement chunked uploads for createIndex, add new chunkedUpsert helper and update typebox validation, update unit test coverage, export types

## Problem
In v0 there was a utility called `chunkedUpsert()` which allowed users to upsert vectors in chunks. https://github.com/pinecone-io/pinecone-ts-client/blob/6fbc502f8803c3d155870262b6f3268d1d441fad/src/v0/utils.ts#L58

We want to offer this functionality in v1 as well to ease usage around the 1000 vector limit for upserting, and give end users more control over their upserts.

## Type of Change

## Test Plan


src/data/upsert.ts
```diff
@@ -14,34 +11,81 @@ export const SparseValuesSchema = Type.Object(
...
+
+  try {
+    await Promise.allSettled(chunkedPromises);
+    return;
+  } catch (e) {
+    throw new Error(`Error upserting vectors into index: ${e}`);
+  }
+};
+
+const sliceArrayToChunks = (array: VectorArray, chunkSize: number) => {
```
jhamon, line 87: Since this code is dense and mathy, I think it's worth exporting this function (locally only, not from the published module) so you can test the chunk logic standalone. I'd want to verify we're not accidentally off-by-1 for various cases and going to accidentally leave behind a vector that should've been upserted (or issue include a vector in multiple upserts). By testing it separately, you can test with simpler arrays without the need to setup entire vector objects.

Potential cases to cover:
- array length 0
- array length 1
- array length even
- array length odd
- array length < chunkSize
- array length > chunkSize
- array divides evenly into chunkSize pieces
- array does not divide evenly into chunkSize pieces
- etc

src/data/upsert.ts
```diff
@@ -14,34 +11,81 @@ export const SparseValuesSchema = Type.Object(
...
+  chunkSize = 10
+) => {
+  // Split vectors into chunks and promises for upserting
+  const chunks = sliceArrayToChunks(vectors, chunkSize);
+
+  const chunkedPromises = chunks.map(async (chunk) => {
+    try {
+      await api.upsert({ upsertRequest: { vectors: chunk, namespace } });
+    } catch (e) {
+      throw new Error(`Error upserting chunk: ${e}`);
```
jhamon, line 75: You will run into this when you try to rebase over the init changes. But I think you could use the `handleApiError` function here that I just merged in to standardize the error handling.

```
const err = await handleApiError(e);
throw err;
```

If you want to prefix the message with extra info you could do something like this by passing a second param.

```
const err = await handleApiError(e, async (statusCode, rawMessage) => (`Error upserting chunk. ${rawMessage}`));
throw err;
```

jhamon (approved):
The general approach looks good to me, but I can tell it's going to be a big pain resolving merge conflicts in this PR. Sorry about that 😬 

Didn't do that much manual testing since I'm thinking we'll want to do that after you rebase and resolve conflicts.

src/data/upsert.ts
austin-denoble: I believe I've covered all these cases along with `batchSize = 0`, which shouldn't occur via our code but figured I'd account for it just to be safe.

Great callout, let me know if there are other scenarios that should be expressed via unit tests.

PR #95, opened 2023-09-04
[v1] List operations return objects instead of strings

jhamon:
## Problem

The `listIndex` and `listCollection` operations currently return arrays of strings. This makes it impossible to add additional information from future versions of the API without it being a breaking change.

## Solution

Map string responses into object responses. By returning objects, we can add in additional properties in the future without it being a breaking change.

## Type of Change

## Test Plan

Updated unit tests.

`npm run repl`

```node
> const pinecone = new Pinecone()
undefined
> await pinecone.listIndexes()
[
  { name: 'delete-me' },
  { name: 'from-collection' },
  { name: 'jen2' },
  { name: 'jen3' }
]
> await pinecone.listCollections()
[
  { name: 'collection3' },
  { name: 'collection1' },
  { name: 'collection-from-jen3' }
]
```

### DIFF
--- a/src/control/__tests__/listCollections.test.ts
+++ b/src/control/__tests__/listCollections.test.ts
@@ -5,7 +5,7 @@ import {
 } from '../../errors';
 
 describe('listCollections', () => {
-  test('should return a list of collections', async () => {
+  test('should return a list of collection objects', async () => {
     const IOA = {
       listCollections: jest
         .fn()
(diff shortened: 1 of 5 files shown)


gdj0nes (approved):

austin-denoble (approved):
🚢 

Nice! Callout for the in the future: we may want to build centralized normalization utilities for our different entities, especially if they'll be changing moving forward.

PR #96, opened 2023-09-04
[v1] Migrate vector/record terminology and add missing type exports

jhamon:
## Problem

1) We have adopted the term `Record` in our [documentation](https://docs.pinecone.io/docs/overview#pinecone-indexes-store-records-with-vector-data) to refer to items stored in an index. 
2) Several type exports were missing from the module.

## Solution

- Eliminate duplication of some type definitions by moving definition for shared concepts such as `Record` and `Index` into `/types.ts` files that can be referenced from multiple actions. 
- Rename the references to `Vector` to `PineconeRecord` everywhere it makes sense. Alias values to the name `vectors` when interacting with the generated client. I chose to avoid using the name `Record` for the type because there is already a [typescript utility with this name](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) and it seems to confuse the typescript compiler and lead to confusing hints & error messages.
- Add module exports for every client methods argument and return types. Some of these types such as `DeleteOneOptions` are just aliases of other types, but giving them a method-specific name affords some flexibility for refactoring in the future if we decide to extend the method to accept other things without introducing a breaking type change.

New files:
- `src/data/types.ts`
- `src/control/types.ts`

## Type of Change

## Test Plan

Only types and exported types changed in this diff. Tests should be passing with changes only to import paths.


src/control/index.ts
```diff
@@ -1,19 +1,34 @@
...
-export { createIndex } from './createIndex';
-export { deleteIndex } from './deleteIndex';
-export { describeIndex } from './describeIndex';
-export { listIndexes } from './listIndexes';
-
-export type { ConfigureIndexOptions } from './configureIndex';
-export type { CreateIndexOptions } from './createIndex';
-export type { IndexList, IndexNameObj } from './listIndexes';
-export type { IndexName } from './describeIndex';
+export { configureIndex, type ConfigureIndexOptions } from './configureIndex';
```
austin-denoble, line 2: **praise:** Appreciate the cleanup consolidating all these. 👍

src/data/types.ts
```diff
@@ -0,0 +1,35 @@
...
+    values: Type.Array(Type.Number()),
+  },
+  { additionalProperties: false }
+);
+export const PineconeRecordSchema = Type.Object(
+  {
+    id: RecordIdSchema,
+    values: RecordValuesSchema,
+    sparseValues: Type.Optional(RecordSparseValuesSchema),
+    metadata: Type.Optional(Type.Object({}, { additionalProperties: true })),
```
austin-denoble, line 27: **question:** Thoughts on `metadata` having it's own type along with the other fields?

src/data/update.ts
```diff
@@ -2,32 +2,29 @@ import { handleApiError } from '../errors';
...
+    id: RecordIdSchema,
+    values: Type.Optional(RecordValuesSchema),
+    sparseValues: Type.Optional(RecordSparseValuesSchema),
     metadata: Type.Optional(Type.Object({}, { additionalProperties: true })),
   },
   { additionalProperties: false }
 );
 
-export type UpdateVectorOptions = Static<typeof UpdateVectorOptionsSchema>;
+export type UpdateVectorOptions = Static<typeof UpdateRecordOptionsSchema>;
```
austin-denoble, line 21: **question:** Is this a case of aliasing when interacting with the OpenAPI client, or should the exported type be `UpdateRecordOptions`?

src/index.ts
```diff
@@ -1,20 +1,41 @@
-export { Pinecone, type PineconeConfiguration } from './pinecone';
+export { Pinecone } from './pinecone';
+export { Index } from './data';
 
 export * as Errors from './errors';
 
 // Type exports
-export type { CreateIndexOptions, IndexList, IndexName } from './control';
 export type {
+  ConfigureIndexOptions,
```
austin-denoble, line 8: **suggestion:** Maybe alphabetize these really chunky export lists? I'm not sure how far we want to go with alphabetizing, but I feel like it can be really helpful for bigger lists of attributes like this where we don't want to miss anything.

src/data/query.ts
```diff
@@ -1,38 +1,41 @@
...
-const QueryByVectorId = Type.Object({
+const QueryByRecordId = Type.Object({
   ...shared,
-  id: nonemptyString,
+  id: RecordIdSchema,
   vector: Type.Optional(Type.Never()),
   sparseVector: Type.Optional(Type.Never()),
 });
 
 const QueryByVectorValues = Type.Object({
```
austin-denoble, line 26: **question:** Should this be `QueryByRecordValues`?

austin-denoble (approved):
Good stuff, really liking the addition of the `/types.ts` files to centralize things. Have a few nits, suggestions, and questions. Overall though good to go. 🚢

src/data/query.ts
jhamon: Yes, good catch.

jhamon: Actually no. In this case, when querying the caller is supplying a list of embedding values (a "vector") they've created using an embedding model. They're not associated with values stored in the index.

PR #97, opened 2023-09-05
[v1] Implement typescript generics for metadata

jhamon:
## Problem

Even with exported typescript types, results from the pinecone client have to be cast by the caller because the metadata property is a flexible data field and is currently untyped. We should allow the caller to tell us what type to expect so they don't have to be repeatedly casting as they work with this data.

## Solution

I have converted the `upsert` and `fetch` actions to a new class-based implementation that allows us to handle the generic Metadata type. The functional changes are minimal (just shuffling around existing logic), but using a class here instead of a high-order-function let me coax typescript generics into doing what I want.

Data returned should be unchanged, but the typescript compiler should understand what data is expected by `upsert` and returned by `fetch` due to the `MovieMetadata` type getting passed by the caller.

```typescript
const pinecone = new Pinecone()

type MovieMetadata = {
  title: string,
  genre: string,
  runtime: number
}

// This is the main usage change. You now pass your expected metadata type 
// as a generic type param.
const index = pinecone.index<MovieMetadata>('movie-embeddings')

const response = index.fetch(['1234'])
const movieRecord = response.vectors['1234']
if (movieRecord && movieRecord.metadata) {
  // Benefit 1: typescript should no longer require any casting to interact with this metadata
  console.log(movieRecord.metadata.genre)
}

// Benefit 2: if you pass the wrong metadata into upsert you should now get typescript feedback
await index.upsert([{ 
  id: '2345', 
  values: [0.2, 0.3, 0.4], 
  metadata: { description: 'this prop not in MovieMetadata' } // should be ts error
}])
```

## Type of Change

## Test Plan

I added unit tests for the typescript errors under `src/data/__tests__/index.test.ts`

Also did more manual testing to confirm upsert and fetch still functional.

## Actions

Eventually we should have all the actions using the same pattern, but on the data plane we only __have__ to do it for those actions which involve metadata in either the method params or in the returned data. That means these are the must-dos:

- [x] upsert
- [x] fetch
- [x] query
- [x] update

For now we don't need to worry about these because they don't input or output metadata:
- deleteOne
- deleteMany
- deleteAll
- describeIndexStats

### DIFF
--- a/src/data/__tests__/index.test.ts
+++ b/src/data/__tests__/index.test.ts
@@ -1,0 +1,80 @@
+import { UpsertCommand } from '../upsert';
+import { FetchCommand } from '../fetch';
+import { Index } from '../index';
+
+jest.mock('../upsert');
+jest.mock('../fetch');
+
+describe('Index', () => {
+  let config;
+
+  beforeEach(() => {
+    config = {
+      apiKey: 'test-api-key',
+      environment: 'test-environment',
+    };
+  });
+
+  test('upsert: has type errors when passing malformed metadata', async () => {
+    type MovieMetadata = {
+      genre: string;
+      runtime: number;
+    };
+    const index = new Index<MovieMetadata>('index-name', config, 'namespace');
+    expect(UpsertCommand).toHaveBeenCalledTimes(1);
+
+    // No ts errors when upserting with propert MovieMetadata
+    await index.upsert([
+      {
+        id: '1',
+        values: [0.1, 0.1, 0.1],
+        metadata: {
+          genre: 'romance',
+          runtime: 120,
+        },
+      },
+    ]);
+
+    // No ts errors when upserting with no metadata
+    await index.upsert([
+      {
+        id: '2',
+        values: [0.1, 0.1, 0.1],
+      },
+    ]);
+
+    // ts error expected when passing metadata that doesn't match MovieMetadata
+    await index.upsert([
+      {
+        id: '3',
+        values: [0.1, 0.1, 0.1],
+        metadata: {
+          // @ts-expect-error
+          somethingElse: 'foo',
+        },
+      },
+    ]);
+  });
+
+  test('fetch: response is typed with generic metadata type', async () => {
+    type MovieMetadata = {
+      genre: string;
+      runtime: number;
+    };
+    const index = new Index<MovieMetadata>('index-name', config, 'namespace');
+    expect(FetchCommand).toHaveBeenCalledTimes(1);
+
+    const response = await index.fetch(['1']);
+    if (response && response.vectors) {
+      // eslint-disable-next-line
+      Object.entries(response.vectors).forEach(([key, value]) => {
+        // No errors on these because they are properties from MovieMetadata
+        console.log(value.metadata?.genre);
+        console.log(value.metadata?.runtime);
+
+        // @ts-expect-error because result is expecting metadata to be MovieMetadata
+        console.log(value.metadata?.bogus);
+      });
+    }
+  });
+});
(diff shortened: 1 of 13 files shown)


src/data/__tests__/index.test.ts
```diff
@@ -0,0 +1,80 @@
+import { UpsertCommand } from '../upsert';
+import { FetchCommand } from '../fetch';
+import { Index } from '../index';
+
+jest.mock('../upsert');
+jest.mock('../fetch');
+
+describe('Index', () => {
```
jhamon, line 8: All the actual functionality of the index actions is tested in more detail elsewhere. I added these tests to see whether the generic type param was resulting in the expected typescript errors. Expected errors are marked `// @ts-expect-error` and the compiler errors if no exception occurs.

src/data/types.ts
```diff
@@ -32,4 +32,14 @@ export const PineconeRecordSchema = Type.Object(
 export type RecordId = Static<typeof RecordIdSchema>;
 export type Values = Static<typeof RecordValuesSchema>;
 export type SparseValues = Static<typeof RecordSparseValuesSchema>;
-export type PineconeRecord = Static<typeof PineconeRecordSchema>;
+export type RecordValues = Array<number>;
+export type RecordSparseValuse = {
+  indices: Array<number>;
+  values: Array<number>;
+};
+export type PineconeRecord<T> = {
```
damargulis, line 40: I think we can make this more specific -- our metadata can't be nested and can only have values of string, boolean, number, or array of strings

`PineconeRecord<T extends Record<string, string | boolean | number | array<string>>`

(might want to make a type for `type MetadataValue = string | boolean | number | array<string>`)

I've done some similar stuff for both metadata values and the filter here: https://github.com/pinecone-io/management-console/blob/main/apps/console/src/types/metadata/language.ts#L1

jhamon: Thanks, I'll give this a shot.

src/data/index.ts
```diff
@@ -37,12 +44,17 @@ export class Index {
...
   deleteOne: ReturnType<typeof deleteOne>;
   describeIndexStats: ReturnType<typeof describeIndexStats>;
-  fetch: ReturnType<typeof fetch>;
   query: ReturnType<typeof query>;
   update: ReturnType<typeof update>;
-  upsert: ReturnType<typeof upsert>;
 
-  constructor(indexName: string, config: ApiConfig, namespace = '') {
+  private fetchCommand: FetchCommand<T>;
+  private upsertCommand: UpsertCommand<T>;
```
austin-denoble, line 51: **question:** Should the other index functions here also be marked private?

src/data/__tests__/index.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  test('upsert: has type errors when passing malformed metadata', async () => {
+    type MovieMetadata = {
+      genre: string;
+      runtime: number;
+    };
+    const index = new Index<MovieMetadata>('index-name', config, 'namespace');
+    expect(UpsertCommand).toHaveBeenCalledTimes(1);
+
+    // No ts errors when upserting with propert MovieMetadata
```
austin-denoble, line 26: **tiny nit:** `"propert"` -> `"proper"`

austin-denoble (approved):
Approach looks good. Also really appreciate Daniel's point around making the metadata type a bit more specific to our purposes. Great work, I feel like this will be really helpful for consumers. Rebase should resolve all the typing refactor changes which are now represented here I'm assuming.

iamannamai (commented):
Retroactive comment, because I didn't get to this, but nice work here! As always, really appreciate the self-documenting nature of the tests, I always rely on them to understand what's happening in this code.

One suggestion/thought--Vitest does this out of the box, but if it's helpful for future use cases, we can have type tests for this project if the `console.logs` stop serving us: https://www.npmjs.com/package/ts-expect

I haven't done a lot of type testing, and this particular package is not widely used enough for me to recommend this definitively, but possibly worth looking at!

PR #104, opened 2023-09-10
fix docs query exemple

glody007:
In this case, the query function should be used on the namespace and on the index

## Type of Change

- [ ] Non-code change (docs)


jhamon:
Nice catch, thanks.

PR #107, opened 2023-09-11
Remove typebox Static<> type inference in favor of TypeScript

austin-denoble:
## Problem
We use `typebox` for runtime validation of arguments in the Node client. Currently, we're leveraging typebox's `Static<typeof schema>` utility for generating TypeScript types out of these runtime schemas. 

We've noticed this doesn't play well with documentation generation, and there have been some bug reports regarding issues  possibly related to typebox's generated types.

## Solution
Instead of relying on the typebox utility, use our own TypeScript types which are manually aligned with existing schemas.

## Type of Change

## Test Plan
These are all typing changes so there shouldn't be any client code affected by these changes.

Make sure to run:
`npm run lint`

### DIFF
--- a/src/data/deleteMany.ts
+++ b/src/data/deleteMany.ts
@@ -1,7 +1,7 @@
 import { VectorOperationsProvider } from './vectorOperationsProvider';
 import { handleApiError } from '../errors';
 import { buildConfigValidator } from '../validator';
-import { Static, Type } from '@sinclair/typebox';
+import { Type } from '@sinclair/typebox';
 import type { DeleteRequest } from '../pinecone-generated-ts-fetch/models/DeleteRequest';
 import { RecordIdSchema } from './types';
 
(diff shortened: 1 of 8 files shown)


src/data/deleteMany.ts
```diff
@@ -17,11 +17,11 @@ const DeleteManySchema = Type.Union([
...
 
-export type DeleteManyByVectorIdOptions = Static<
-  typeof DeleteManyByRecordIdSchema
->;
-export type DeleteManyByFilterOptions = Static<typeof DeleteManyByFilterSchema>;
-export type DeleteManyOptions = Static<typeof DeleteManySchema>;
+export type DeleteManyByVectorIdOptions = Array<string>;
+export type DeleteManyByFilterOptions = object;
+export type DeleteManyOptions =
+  | DeleteManyByVectorIdOptions
```
jhamon, line 23: I've never seen the `|` used on the first entry like this. Is that correct/valid syntax?

jhamon (dismissed):

src/data/deleteMany.ts
austin-denoble: This was a prettier addition, but yeah seems like valid syntax. I did have to look it up though, found this: https://github.com/prettier/prettier/issues/6299

jhamon (approved):

PR #109, opened 2023-09-12
Update integration test setup

jhamon:
## Problem

I want Typescript to work properly within my integration tests without having to manage a separate tsconfig file.

## Solution

Getting this working was a bit of an adventure. As usual, when I thought I was 80% done I was only actually about 20% done. First I did these things:

- Move integration tests underneath the `src/` directory, which is the `rootDir` in our `tsconfig.json` file.
- Update test commands in package.json so that `npm run test:unit` does not include integration tests.
- Add a new test, `src/integration/data/deleteMany.test.ts`, using a randomly generated namespace to store test-specific state in the index from other test runs. In testing these data plane operations, it seems fine to recycle the same index and save the time it would take to spin up a fresh one for every test.

That was enough to get tests passing locally with `npm run test:integration`. Then I tried running them in CI and everything was mysteriously failing. After some investigation, I realized there was no `fetch` present in the node environment where things run on Github Actions. I'm still somewhat baffled how this was working on my local machine, in the node repl and in the sample apps, but not in this node CI env.

To address this:
- I explicitly import `cross-fetch` and pass that fetch implementation to the openapi client. Node doesn't natively have a `fetch` command. `cross-fetch` ([npm](https://www.npmjs.com/package/cross-fetch)) is a polyfill proxy that attempts to activate the correct fetch polyfill for the current environment.
- Updated tests for the projectIdSingleton, which is the only thing I've written in this client that used fetch directly. All other unit tests are mocking things at the openapi object layer so had no direct dependency on fetch.

Also:
- Cleanup and remove vestiges of the old integration test suite, which hasn't been running for quite a while.

## Type of Change

## Test Plan

CI should be green.


src/integration/test-helpers.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    vectors.push(vector);
+  }
+  return vectors;
+};
+
+export const generateSparseValues = (dimension: number): RecordSparseValues => {
+  const values: number[] = [];
+  const indecies: number[] = [];
```
austin-denoble, line 66: **nit:** `"indecies"` -> `"indices"`?

src/integration/test-helpers.ts
```diff
@@ -0,0 +1,76 @@
...
+  for (let j = 0; j < dimension; j++) {
+    values.push(Math.random());
+    indecies.push(j);
+  }
+  const sparseValues: RecordSparseValues = {
+    indices: indecies,
+    values: values,
+  };
```
austin-denoble, line 74: **nit:** Any reason to use the variables directly rather than mapping? 

```suggestion
  const sparseValues: RecordSparseValues = {
    indices,
    values
  };
```

src/integration/data/deleteMany.test.ts
```diff
@@ -0,0 +1,89 @@
+import { Pinecone, Index } from '../../index';
+import {
+  randomString,
+  createIndexIfDoesNotExist,
+  generateRecords,
+} from '../test-helpers';
+
+describe('deleteMany', () => {
+  const INDEX_NAME = 'ts-integration';
+  let pinecone: Pinecone, index: Index, ns: Index, namespace: string;
```
austin-denoble, line 10: **nit / suggestion:** Really minor, but maybe tweak the variable names here for readability? 

```suggestion
  let pinecone: Pinecone, index: Index, namespace: Index, namespaceStr: string;
```

src/integration/data/deleteMany.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    const fetchResult = await ns.fetch(['0']);
+    const records = fetchResult.records;
+    if (records) {
+      expect(records['0'].id).toEqual('0');
+      expect(records['0'].values.length).toEqual(5);
+    } else {
+      fail(
+        'Did not find expected records. Fetch result was ' +
+          JSON.stringify(fetchResult)
+      );
```
austin-denoble, line 49: **praise:** Nice idea to include a sort of trace with the failure messages. 👍

src/integration/data/deleteMany.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (stats.namespaces) {
+      expect(stats.namespaces[namespace].recordCount).toEqual(1);
+    } else {
+      fail(
+        'Expected namespaces to be defined (second call). Stats were ' +
+          JSON.stringify(stats)
+      );
+    }
+
+    // Check that record 1 still exists
```
austin-denoble, line 64: **nit:** Maybe off-by-one in my head, but this should should maybe be "record 2" to reflect the failure response and variable naming here.

src/data/__tests__/projectIdSingleton.test.ts
```diff
@@ -7,12 +15,10 @@ describe('ProjectIdSingleton', () => {
 
   test('issues whoami requests for unknown projectId', async () => {
     // @ts-ignore
```
austin-denoble, line 17: **question:** Do we still need the `@ts-ignore` on these since we're no longer overriding `global`?

src/errors/config.ts
```diff
@@ -30,7 +30,7 @@ export class PineconeEnvironmentVarsNotSupportedError extends BasePineconeError
 export class PineconeUnknownRequestFailure extends BasePineconeError {
   constructor(url: string, underlyingError: Error) {
     const message = `Something went wrong while attempting to call ${url}. Please check your configuration and try again later. Underlying error was ${JSON.stringify(
-      underlyingError
+      underlyingError.message
```
austin-denoble, line 33: **praise:** Nice catch (assuming from debugging some of the CI issues)

austin-denoble (approved):
Nice work, thanks for taking on the integration test fixes really quickly. Just some nits and questions but this looks good to go. 🚢

src/integration/test-helpers.ts
jhamon: No reason, I copied this util out of an old test helper without much scrutiny. I agree with your suggestions.

src/integration/data/deleteMany.test.ts
jhamon: I made some adjustments to make things more clear. Since the records we upsert are generated with a helper, you can't easily see what the ids are. So I added some expects to confirm the three records have ids 0, 1, and 2 before moving on to deleting things.

src/data/__tests__/projectIdSingleton.test.ts
jhamon: We do because typescript doesn't seem to understand that the real implementation got replaced with a mock and `mockResolvedValue` isn't part of the real fetch interface.

src/errors/config.ts
jhamon: Yep.... 💀

Issue #108, opened 2023-09-12
[Bug] A Node.js API is used (fs.readFileSync) which is not supported in the Edge Runtime 


jaukia:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Getting error like this:

[0] - error ./node_modules/@pinecone-database/pinecone/dist/utils/user-agent.js:27:22
[0] Module not found: Can't resolve 'fs'
[0] 
[0] https://nextjs.org/docs/messages/module-not-found
[0] 
[0] Import trace for requested module:
[0] ./node_modules/@pinecone-database/pinecone/dist/utils/index.js
[0] ./node_modules/@pinecone-database/pinecone/dist/pinecone.js
[0] ./node_modules/@pinecone-database/pinecone/dist/index.js

Looks like user-agent assumes fs exists.

### Expected Behavior

Pinecone would work even on edge

### Steps To Reproduce

Call Pinecone from node:

const pinecone = new Pinecone({
      apiKey: process.env.PINECONE_API_KEY,
      environment: process.env.PINECONE_ENVIRONMENT
    });
const pineconeIndex = pinecone.index(process.env.PINECONE_INDEX).namespace('ns1');

...

const value = await index.query({
    topK: topK,
    vector: embeddingResponse.data[0].embedding,
    filter: filter,
  });

### Environment

```markdown
- **OS**: Mac
- **Language version**: TypeScript
- **Pinecone client version**: 1.0.0
```

jaukia:
Trying to downgrade to 0.1.6. gives me this error:

[0]  [PineconeError: PineconeClient: Error calling query: PineconeError: PineconeClient: Error calling queryRaw: FetchError: The request failed and the interceptors did not return an alternative response] {
[0]   name: 'PineconeError'

Has the API changed or am I doing something wrong?

RickRyan26:
Same problem. The only reason im using Pinecone in particular is for the Edge.

jhamon:
Sorry for the trouble, I'll try to get a fix for this up soon.

jhamon:
[...]

PR #110, opened 2023-09-13
Add support for Edge Runtime

jhamon:
## Problem

People attempting to use this library in Vercel's [Edge Runtime](https://edge-runtime.vercel.app/) are hitting runtime exceptions due to reliance on some node utilities that are not available.

This addresses issue #108 

## Solution

- Setup new `npm run test:integration:edge` which runs integration tests using a special jest context designed to highlight when code is using dependencies that are not available on edge.
- Integrate this test run into CI

Code changes to get these tests passing:
- Add an `isEdge()` utility that looks for a global var, `EdgeRuntime` to detect whether execution is happening in the Edge Runtime.
- Use that util to guard things which won't work in Edge. Things not supported in Edge:
  - Pulling `os` info for the user-agent string. This info can just be omitted from the user-agent.
  - Using `fs` to read `package.json` version, also for the user-agent string. The version number is the most important thing in the user-agent so I do need some other way of getting that info. Importing json as a module is possible in es6, but I was getting typescript build errors when trying to grab package.json directly because that file is outside of the tsconfig `rootDir`. So, I adjusted the publish steps to dump version info to a place it can be safely referenced, `src/version.json`.
  - Runtime validation using AJV. The code generation that happens in the `compile` step causes Edge to blow up. So we skip runtime validations for Edge for now. If they are developing with TypeScript, these should be largely unnecessary anyway.

## Type of Change

## Test Plan

- New CI job for edge integration tests should be green.
- For changes to publish jobs, I did some local testing to explore what the npm version command does and test the jq commands. But won't know for sure that this is all correct,  though, until trying to ship a release.

### DIFF
--- a/.github/actions/npm-release/action.yml
+++ b/.github/actions/npm-release/action.yml
@@ -25,9 +25,16 @@ runs:
       run: |
         git config --global user.email "${{ inputs.git_email }}"
         git config --global user.name "${{ inputs.git_username }}"
-    - name: 'Bump version and commit'
+    - name: 'Bump version'
       shell: bash
-      run: npm version ${{ inputs.releaseType }} -m '[skip ci] Publish release %s'
+      run: npm version ${{ inputs.releaseType }} --no-git-tag-version
+    - name: 'Commit and tag version changes'
+      shell: bash
+      run: |
+        git add .
+        newVersion=$(jq -r '.version' package.json)
+        git commit -m "[skip ci] Publish release v$newVersion"
+        git tag "v$newVersion"
     - name: 'Output release tag'
       shell: bash
       id: release-tag-step
--- a/.github/workflows/release-dev.yml
+++ b/.github/workflows/release-dev.yml
@@ -35,6 +35,7 @@ jobs:
           jq --arg newVersion "$devVersion" '.version = $newVersion' package.json > package.tmp && mv package.tmp package.json
           jq --arg newVersion "$devVersion" '.version = $newVersion' package-lock.json > package-lock.tmp && mv package-lock.tmp package-lock.json
           jq --arg newVersion "$devVersion" '.packages[""].version = $newVersion' package-lock.json > package-lock.tmp && mv package-lock.tmp package-lock.json
+          jq --null-input --arg version "$devVersion" '{"name": "@pinecone-database/pinecone", "version": $version}' > src/version.json
 
       - name: 'Publish to npm'
         run: npm publish --tag next
--- a/src/validator.ts
+++ b/src/validator.ts
@@ -1,6 +1,7 @@
 import Ajv from 'ajv';
 import type { ErrorObject } from 'ajv';
 import { PineconeArgumentError } from './errors';
+import { isEdge } from './utils/environment';
 
 const prepend = (prefix: string, message: string) => {
   return `${prefix} ${message}`;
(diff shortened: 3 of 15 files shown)


.github/actions/npm-release/action.yml
```diff
@@ -25,9 +25,16 @@ runs:
       run: |
         git config --global user.email "${{ inputs.git_email }}"
         git config --global user.name "${{ inputs.git_username }}"
-    - name: 'Bump version and commit'
+    - name: 'Bump version'
       shell: bash
-      run: npm version ${{ inputs.releaseType }} -m '[skip ci] Publish release %s'
+      run: npm version ${{ inputs.releaseType }} --no-git-tag-version
+    - name: 'Commit and tag version changes'
```
jhamon, line 31: These changes plus the addition of the `version` script in the `package.json` file are needed to update `src/version.json` with the new version number prior to the commit/tag. Even though the `npm version` command automatically invokes the `npm run version` script for us, it doesn't happen until _after_ the automatic commit. So we need to tell npm not to do the auto commit with the `--no-git-tag-version` flag and handle it ourselves.

.github/workflows/release-dev.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           jq --arg newVersion "$devVersion" '.version = $newVersion' package.json > package.tmp && mv package.tmp package.json
           jq --arg newVersion "$devVersion" '.version = $newVersion' package-lock.json > package-lock.tmp && mv package-lock.tmp package-lock.json
           jq --arg newVersion "$devVersion" '.packages[""].version = $newVersion' package-lock.json > package-lock.tmp && mv package-lock.tmp package-lock.json
+          jq --null-input --arg version "$devVersion" '{"name": "@pinecone-database/pinecone", "version": $version}' > src/version.json
```
jhamon, line 38: For dev builds, we never relied on the `npm version` command to set version numbers. So we update `src/version.json` a little differently also. I tested this jq command on my local machine.

src/validator.ts
```diff
@@ -224,6 +225,11 @@ export const errorFormatter = (subject: string, errors: Array<ErrorObject>) => {
 };
 
 export const buildValidator = (errorMessagePrefix: string, schema: any) => {
+  if (isEdge()) {
```
jhamon, line 228: This turns runtime validation into a noop in the Edge Runtime.

.github/actions/npm-release/action.yml
```diff
@@ -25,9 +25,16 @@ runs:
       run: |
         git config --global user.email "${{ inputs.git_email }}"
         git config --global user.name "${{ inputs.git_username }}"
-    - name: 'Bump version and commit'
+    - name: 'Bump version'
       shell: bash
-      run: npm version ${{ inputs.releaseType }} -m '[skip ci] Publish release %s'
+      run: npm version ${{ inputs.releaseType }} --no-git-tag-version
```
austin-denoble, line 30: I think we need `run` here?

```suggestion
      run: npm run version ${{ inputs.releaseType }} --no-git-tag-version
```

austin-denoble (approved):
LGTM, thanks for getting on top of this so quickly. One callout in the `/npm-release/action.yml` file that I think needs updating.

.github/actions/npm-release/action.yml
jhamon: There is actually a command called [`npm version`](https://docs.npmjs.com/cli/v9/commands/npm-version). It knows how to bump versions and whatnot and commit those changes (unless you disabled that with a flag as we have here).

As part of that command run, it executes certain [lifecycle scripts](https://docs.npmjs.com/cli/v10/using-npm/scripts#npm-version). So if you have defined a `version` command in the `scripts` section of your `package.json`, it will automatically be invoked. There's also a `preversion` hook. But you don't need to call `npm run version` yourself. I agree it's confusing that the command name and lifecycle script hooks have the same name.

austin-denoble: Ah, good to know. Thank you!

PR #106, opened 2023-09-11
Add build-docs action for validating documentation build during pr workflow

austin-denoble:
## Problem
Currently, we don't have any CI/CD processes for handling documentation and everything is rolled by hand. We're beginning work to automate some of our client documentation to ease the pain around keeping documentation in sync with code.

See the Spike document for more context: https://docs.google.com/document/d/1pz1BKe0NYBcQda5Dp6--D9Cf7Hc3-eXvhjNpONde89U/edit?usp=sharing

## Solution
For now keep changes small. We're moving forward leveraging TypeDoc as our generator as we're already using it as a dependency, and it's easy to integrate quickly with GitHub pages.

- Add a new action (`.github/actions/build-docs/action.yml`) to handle triggering the typedoc build in CI. For now we just want to run and validate that nothing broke.
- Update `.github/workflows/pr.yml` to include the `build-docs` action as a part of `basic-hygiene`. This should run the docs build on PRs, updates, and pushes to main along with prettier and eslint.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [X] Infrastructure change (CI configs, etc)
- [ ] Non-code change (docs, etc)
- [ ] None of the above: (explain here)

## Test Plan
We want to validate that the action is running as expected and not causing issues in CI. You can validate the added step in this PR:

- For this PR, check the workflow and make sure the `"Linting, formatting, etc"` step includes the TypeDoc step and it completes as expected: https://github.com/pinecone-io/pinecone-ts-client/actions/runs/6151814963/job/16692664139?pr=106 Screenshot 2023-09-11 at 5 11 35 PM 
- More generally, for other PRs after this lands - navigate to https://github.com/pinecone-io/pinecone-ts-client/actions/workflows/pr.yml to validate that the pr workflow run is completing as expected

PR #112, opened 2023-09-13
Add merge .github workflow to build and deploy docs

austin-denoble:
## Problem
We're working to improve documentation for the SDK clients by automating the process of generating and deploying documentation. Specifically, this needs to happen on every merge to `main`.

## Solution

- Create a new `.github/workflows/merge.yml` file that will be triggered specifically on closed PRs that are merged to `main`. For now this explicitly handles the documentation pieces, but more steps could be added in the future if there are other CI processes we want to trigger on merge.
- Reuse the `build-docs` action setup in the previous PR (#106) for generation.
- Use the `s0/git-publish-subdir-action` to push the documentation artifact to a specific branch that will house current client documentation: `gh-pages`.

**Note:** I went back and forth on either adding a new workflow explicitly for merging, or hooking into the existing `.github/workflows/pr.yml` and using `if: github.event.pull_request.merged == true` to split the deployment behavior out. I'm still open to this solution if we'd prefer to keep it simple.

It also feels like we're running the `build-docs` step a lot across the `pr.yml` and now `merge.yml` workflows. Maybe this is ok since we're only using the output in merge, but wanted to call it out in case there's some work that could be done there passing things around between jobs, etc.

## Type of Change

## Test Plan
Make sure CI is green.

After this merges I can make the adjustments to the repository settings to pull & publish from `gh-pages`, and we can validate the docs are built correctly at https://pinecone-io.github.io/pinecone-ts-client.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -1,14 +1,13 @@
 name: 'Merge to main'
 
 on:
+  push:
+    branches:
+      - main
   workflow_dispatch: {}
-  pull_request:
-    types:
-      - closed
 
 jobs:
   build-and-deploy-documentation:
-    if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -1,0 +1,30 @@
+name: 'Merge to main'
+
+on:
+  workflow_dispatch: {}
+  pull_request:
+    types:
+      - closed
+  workflow_call:
+    secrets:
+      GITHUB_TOKEN:
+        required: true
+
+jobs:
+  build-and-deploy-documentation:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup
+        uses: ./.github/actions/setup
+      - name: Generate TypeDoc documentation
+        uses: ./.github/actions/build-docs
+      - name: Deploy documentation to gh-pages
+        uses: s0/git-publish-subdir-action@develop
+        env:
+          REPO: self
+          BRANCH: gh-pages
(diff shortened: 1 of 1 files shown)


.github/workflows/merge.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup
+        uses: ./.github/actions/setup
+      - name: Generate TypeDoc documentation
+        uses: ./.github/actions/build-docs
+      - name: Deploy documentation to gh-pages
+        uses: s0/git-publish-subdir-action@develop
```
austin-denoble, line 21: More information on this action can be found here: https://github.com/marketplace/actions/push-git-subdirectory-as-branch

jhamon:
Any thoughts on this alternative trigger?

```
  push:
    branches:
      - main
 ```

This trigger will run on every commit that makes it to the `main` branch, which is effectively every merged PR + any yolo commits direct to main from admins (rare but they happen). I don't think what you have currently would catch yolo commits. Docs-only changes are the kind of thing we __are__ relatively likely to push without review, so I think it makes sense to consider this case.

Anyway this seems low risk to merge as-is. We can tinker on triggers as we go.

jhamon (approved):

austin-denoble:
> Any thoughts on this alternative trigger?
> 
> ```
>   push:
>     branches:
>       - main
> ```
> 
> This trigger will run on every commit that makes it to the `main` branch, which is effectively every merged PR + any yolo commits direct to main from admins (rare but they happen). I don't think what you have currently would catch yolo commits. Docs-only changes are the kind of thing we **are** relatively likely to push without review, so I think it makes sense to consider this case.
> 
> Anyway this seems low risk to merge as-is. We can tinker on triggers as we go.

Good call, I'll update. Would be good to cover all cases of code making it into main rather than the explicit PR path. 👍

Issue #105, opened 2023-09-10
[Bug] Typescript compilation problems while upgrading pinecone client in angular project


zmiju:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

I installed Pinecone in angular project in firebase functions with npm uninstall @pinecone-database/pinecone.

Previously it worked well, but after introducing version 1 of pinecone, I cannot deploy functions. I receive over 1000 bugs while deploying. Here is the summary of the message:

> Errors Files 51 node_modules/.pnpm/typescript@4.9.3/node_modules/typescript/lib/lib.dom.d.ts:25 1002 node_modules/@types/web/index.d.ts:7

They disappear when I uninstall pinecone again, this is how I know that the problem is related to pinecone.
I even downgraded pinecone library to version 0.1.6 and it started working again.

Exapmple of bugs are: 

> node_modules/@types/web/index.d.ts:18116:13 - error TS2403: Subsequent variable declarations must have the same type. Variable 'XPathResult'
> node_modules/.pnpm/typescript@4.9.3/node_modules/typescript/lib/lib.dom.d.ts:17276:14 17276 readonly FIRST_ORDERED_NODE_TYPE: number;
> node_modules/@types/web/index.d.ts:18111:14 - error TS2717: Subsequent property declarations must have the same type.
> node_modules/@types/web/index.d.ts:17593:13 - error TS2403: Subsequent variable declarations must have the same type.

I reinstalled pinecone several times and nothing worked. I also searched for answer on pinecone forum, but the problem didn't appear. I created totally new angular project, copied sources files and tried to launch functions from scratch, but it didn't help either.

### Expected Behavior

To work well after installing new pinecone version for angular project firebase functions.

### Steps To Reproduce

1. Generate new angular project with functions
2. Go to functions folder
3. Install pinecone with npm i @pinecone-database/pinecone
4. Try to deploy function with firebase deploy --only functions 

### Relevant log output

```shell
../node_modules/@types/web/index.d.ts:17360:14 - error TS2717: Subsequent property declarations must have the same type.  Property 'INT_VEC4' must be of type 'number', but here has type '35669'.

17360     readonly INT_VEC4: 0x8B55;
                   ~~~~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:16407:14
    16407     readonly INT_VEC4: GLenum;
                       ~~~~~~~~
[...]

Issue #102, opened 2023-09-07
[Bug] v1.0.0 transpile errors


GoncaloBastos:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When running `npx tsc` several errors occur.

### Expected Behavior

Should transpile correctly.

### Steps To Reproduce

1. import v1.0.0
2. try to transpile

### Relevant log output

```shell
/@sinclair/typebox/typebox.d.ts:179:167 - error TS2589: Type instantiation is excessively deep and possibly infinite.

179 ] extends [TArray, TNumber] ? AssertType<T['items']> : K extends TTemplateLiteral ? TIndexReduce<T, TTemplateLiteralKeyRest<K>> : K extends TUnion<TLiteral<Key>[]> ? TIndexReduce<T, TUnionLiteralKeyRest<K>> : K extends TLiteral<Key> ? TIndexReduce<T, [K['const']]> : TNever;

[...]

PR #114, opened 2023-09-14
Resolve typescript compilation problems

jhamon:
## Problem

Several users have reported issues with typescript compilation

This PR addresses:
- #102
- #105 

### Clean up unused dependencies to resolve web-related type errors

I started by deleting dependencies and running tests to see what dependencies were no longer in use. I was able to delete these:
- @types/web
- @jest/globals
- unique-names-generator

Removing `@types/web` seems to have resolved the majority of Typescript issues related to conflicts with web types in `lib.dom.d.ts`. Since I was already in there mucking with dependencies, I went ahead and shuffled a few things between dependencies/devDependencies just to align with my current understanding of things. 

### TS1005 (Typescript versions <=4.4)

People on older typescript versions report seeing many errors like this:

```
Error: ../pinecone-ts-client/dist/control/index.d.ts(1,31): error TS1005: ',' expected.
```

These errors were caused by using import statements that combined type and value imports onto one line. This syntax was evidently not supported in versions of typescript prior to 4.4.

Resolving these issues involved refactoring about 30 import statements along these lines:
```diff
-import { IndexNameSchema, type IndexName } from './types';
+import { IndexNameSchema } from './types';
+import type { IndexName } from './types';
```

### Typebox errors (Typescript >= 5.1)

The `@sinclair/typebox` dependency is causing issues with very recent versions of typescript.

```
../node_modules/@sinclair/typebox/typebox.d.ts:179:167 - error TS2589: Type instantiation is excessively deep and possibly infinite.

179 ] extends [TArray, TNumber] ? AssertType<T['items']> : K extends TTemplateLiteral ? TIndexReduce<T, TTemplateLiteralKeyRest<K>> : K extends TUnion<TLiteral<Key>[]> ? TIndexReduce<T, TUnionLiteralKeyRest<K>> : K extends TLiteral<Key> ? TIndexReduce<T, [K['const']]> : TNever;
```

### New CI job: `typescript-compilation-tests`

To verify this compilation issue doesn't come up again, I added a new github actions job along with a version matrix. It tests every major version of typescript back to version 4.1 (released [fall 2020](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/)) along with a version of pinecone, then checks whether it can successfully `tsc`. I ran these with the v1.0.0 pinecone release to see the jobs fail as reported, then updated it to install the latest code in this branch and saw them pass. This will be part of CI for every PR from now on.

## Type of Change

## Test Plan

Should see new compilation tests passing in CI. There are now 13 different checks running in CI to exercise compilation with major versions of TypeScript.

## Follow-up

Probably we should add a lint rule to prevent us adding back any mixed type/value imports.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -67,3 +67,58 @@ jobs:
         env:
           CI: true
         run: npm run test:unit -- --coverage
+
+  typescript-compilation-tests:
+    name: TS compile
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        tsVersion:
+          [
+            '~4.1.0',
+            '~4.2.0',
+            '~4.3.0',
+            '~4.4.0',
+            '~4.5.0',
+            '~4.6.0',
+            '~4.7.0',
+            '~4.8.0',
+            '~4.9.0',
+            '~5.0.0',
+            '~5.1.0',
+            '~5.2.0',
+            'latest',
+          ]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          registry-url: 'https://registry.npmjs.org'
+      - name: Install npm packages
+        run: |
+          npm install --ignore-scripts
+        shell: bash
+      - name: Build typescript for pinecone
+        run: npm run build
+        shell: bash
+      - name: install, compile, and run sample app
+        shell: bash
+        env:
+          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
+          PINECONE_ENVIRONMENT: ${{ secrets.PINECONE_ENVIRONMENT }}
+        run: |
+          set -x
+          cd ..
+          cp -r pinecone-ts-client/ts-compilation-test ts-compilation-test
+          cd ts-compilation-test
+          npm install typescript@${{matrix.tsVersion}} --no-cache
+          npm install '../pinecone-ts-client' --no-cache
+          cat package.json
+          cat package-lock.json
+          npm run tsc-version
+          npm run build
+          npm run start
--- a/package.json
+++ b/package.json
@@ -27,9 +27,8 @@
     "node": ">=14.0.0"
   },
   "devDependencies": {
-    "@jest/globals": "^29.3.1",
+    "@edge-runtime/jest-environment": "^2.3.3",
     "@types/jest": "^29.5.0",
-    "@types/node": "^18.11.17",
     "@typescript-eslint/eslint-plugin": "^5.59.11",
     "@typescript-eslint/parser": "^5.59.11",
     "dotenv": "^16.0.3",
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -38,21 +37,18 @@
...
   "types": "dist/index.d.ts",
   "files": [
     "/dist"
   ],
   "dependencies": {
-    "@edge-runtime/jest-environment": "^2.3.3",
     "@edge-runtime/types": "^2.2.3",
-    "@sinclair/typebox": "^0.28.15",
-    "@types/web": "^0.0.99",
+    "@sinclair/typebox": "^0.29.0",
```
jhamon, line 48: Typebox got a small bump to resolve an error on TS >5.1. I didn't bump all the way to latest because that caused problems with some major releases of TS 4.x.

src/control/index.ts
```diff
@@ -1,34 +1,29 @@
...
-  describeIndex,
-  type DescribeIndexOptions,
-  type IndexDescription,
-} from './describeIndex';
-export {
-  listIndexes,
-  type IndexList,
-  type PartialIndexDescription,
-} from './listIndexes';
+export { configureIndex } from './configureIndex';
```
jhamon, line 2: These formatting changes don't read nearly as nice, but are necessary for compatibility with older versions of typescript.

src/utils/os.ts
```diff
@@ -2,6 +2,8 @@ import * as os from 'os';
 
 export const getOSInfo = () => {
   // If available, capture information about the OS
+
+  // @ts-ignore
```
jhamon, line 6: I'm actually deleting this file in a different PR, but TS 4.1 was upset that I was checking all these things in the if-statement that its knowledge of types told it should always be defined. Regardless of what the type says, i think it's prudent to be cautious regarding things you expect from the environment.

src/v0/index.ts
```diff
@@ -35,10 +35,12 @@ async function streamToArrayBuffer(
       break;
     }
 
-    const newResult = new Uint8Array(result.length + value.length);
-    newResult.set(result);
-    newResult.set(value, result.length);
-    result = newResult;
+    if (value) {
```
jhamon, line 38: Older TS compilers were erroring on the `value.length` reference below because the type says `value` could be undefined. I don't think it's a realistic case you would encounter in runtime, though due to the break on done condition above.

austin-denoble (approved):
Great work getting all these issues resolved so quickly. Overall this looks good, and thanks for walking through the thinking around the approach. I think explicitly covering variations in TS versions will be really helpful for us moving forward, especially given a lot of the feedback coming in post V1. 🚢

.github/workflows/pr.yml
```diff
@@ -67,3 +67,58 @@ jobs:
...
+          PINECONE_ENVIRONMENT: ${{ secrets.PINECONE_ENVIRONMENT }}
+        run: |
+          set -x
+          cd ..
+          cp -r pinecone-ts-client/ts-compilation-test ts-compilation-test
+          cd ts-compilation-test
+          npm install typescript@${{matrix.tsVersion}} --no-cache
+          npm install '../pinecone-ts-client' --no-cache
+          cat package.json
+          cat package-lock.json
```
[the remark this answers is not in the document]
jhamon: Yeah, my thinking was that it would be useful to be able to see what the commands actually installed. To speed debugging also in case of future failiures.

PR #115, opened 2023-09-14
Remove references to os for Edge Runtime compatibility

jhamon:
## Problem

PR #110 resolved the problem with `fs` imports that was happening in the Edge Runtime, but feedback on a dev build shows that `os` imports are still an issue even though the `@edge-runtime/jest-environment` test build is passing.

This relates to:
- #108 

## Solution

For now, the easiest thing is to just remove references to `os`. The information it adds to the user-agent header is nice to have, but not essential.

## Type of Change


austin-denoble (approved):
Looks good. 🚢

PR #116, opened 2023-09-14
Use semantic-search example app as client integration test

jhamon:
## Problem

We want to keep example apps up to date with the latest work happening in the SDKs.

## Solution

One day I would like to automatically open PRs with client version bumps on every example repo after every publish. But a good first step in that direction is that we should run the example app tests using the dev version of the client. And a good first step toward that is to pilot this approach using an example app that already contains tests: [`semantic-search-example`](https://github.com/pinecone-io/semantic-search-example).

If we see a failure here, we'll know very early on that we've introduced a breaking change and that work may be needed to update the example apps when the next release is published. And of course the better the tests are in the example apps, the better the signal is here.

## Type of Change

## Test Plan

See new check `Example app: semantic search` is passing.


austin-denoble (approved):
Great thinking leveraging one of our existing examples for this purpose. Also appreciate learning you can use `npm install '../pinecone-ts-client'` to directly install a dependency. 👍

PR #119, opened 2023-09-14
Update typedoc.json configurations and type exports

austin-denoble:
## Problem
There are some basic configurations for TypeDoc we can provide to clean up some of the documentation output. There were also a number of warnings being output when running the utility calling out internals that are represented in documentation, but not included due to not being available as an export at the top level.

## Solution

- Update `typedoc.json` config to exclude `/v0/` code paths, include the version number in the output, and hide the generator tagline at the bottom of the page.
- Update the `{@link PineconeArgumentError}` doc strings to `{@link Errors.PineconeArgumentError` to properly pull in the references.
- Export missing types: `IndexName`, `CollectionName`, `DeleteManyByVectorIdOptions`, `DeleteManyByFilterOptions`, and `QueryShared`.

## Type of Change

## Test Plan
Make sure CI is green.
Validate documentation after it's been built and deployed: https://pinecone-io.github.io/pinecone-ts-client/

The documentation CI step should also still show warnings for some of the `pinecone-generated-ts-fetch` internals, and a few of the index classes. I'm honestly unsure if those should be exported, and exporting them at the top level pulled in additional pieces into warnings so it may need a more thorough pass.

Current warnings should look like this:
Screenshot 2023-09-14 at 4 39 59 PM

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -26,7 +26,11 @@ export type {
   RecordMetadataValue,
 } from './types';
 export { PineconeConfigurationSchema } from './types';
-export type { DeleteManyOptions } from './deleteMany';
+export type {
+  DeleteManyOptions,
+  DeleteManyByFilterOptions,
+  DeleteManyByVectorIdOptions,
+} from './deleteMany';
 export type { DeleteOneOptions } from './deleteOne';
 export type {
   DescribeIndexStatsOptions,
@@ -41,6 +45,7 @@ export type {
   QueryByVectorValues,
   QueryOptions,
   QueryResponse,
+  QueryShared,
 } from './query';
 
 export class Index<T extends RecordMetadata = RecordMetadata> {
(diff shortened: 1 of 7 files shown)


src/data/index.ts
```diff
@@ -26,7 +26,11 @@ export type {
   RecordMetadataValue,
 } from './types';
 export { PineconeConfigurationSchema } from './types';
-export type { DeleteManyOptions } from './deleteMany';
+export type {
+  DeleteManyOptions,
+  DeleteManyByFilterOptions,
+  DeleteManyByVectorIdOptions,
```
jhamon, line 32: Can we rename `DeleteManyByVectorIdOptions` to `DeleteManyByRecordIdOptions` real quick?

jhamon (approved):

src/data/index.ts
austin-denoble: Yup! Lemme update.

PR #123, opened 2023-09-15
Add docstrings to Index class properties and methods

austin-denoble:
## Problem
Now that we're generating and deploying documentation for the Node SDK, we need to fill in docstrings on the `Index` class properties and methods. My thinking is we should have detailed docstrings for all the top-level entry points for the client, and we're currently missing the data plane.

## Solution
Add docstrings to `/src/data/index.ts`. 
New comments for: the constructor, `deleteAll`, `deleteMany`, `deleteOne`, `describeIndexStats`, `fetch`, `namespace`, `query`, `update`, `upsert`.

### Considerations 
I kept things fairly minimal on this first pass as I wanted to get some feedback. I was cribbing a lot from the `README.md` file, and I wasn't sure how in-depth we wanted to go in terms of inline examples, etc. Definitely open to feedback on how granular we'd like to get in terms of docstring contents.

## Type of Change

## Test Plan
There are no direct code changes in this patch so testing should be confined to making sure the docs-build CI job is green, and we're not seeing a bunch of errors around the docs generation process.

You can checkout the branch locally and run the docs command manually to generate things, and then navigate to `./docs/index.html` to validate the output.

`npm run docs:build`

|Before - Node SDK Index Page|After - Node SDK Index Page|
| --- | --- |
|Screenshot 2023-09-15 at 5 29 59 PM|Screenshot 2023-09-15 at 5 29 44 PM|

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -48,6 +48,9 @@ export type {
   QueryShared,
 } from './query';
 
+/**
+ * @typeParam T - The type of metadata associated with each record.
+ */
 export class Index<T extends RecordMetadata = RecordMetadata> {
   private config: PineconeConfiguration;
   private target: {
@@ -55,9 +58,69 @@ export class Index<T extends RecordMetadata = RecordMetadata> {
     namespace: string;
   };
 
+  /**
+   * Delete all records from the index.
+   *
+   * @example
+   * ```js
+   * const pinecone = new Pinecone();
+   * await pinecone.index('my-index').deleteAll();
+   * ```
+   * @returns A promise that resolves when the delete is completed.
+   */
   deleteAll: ReturnType<typeof deleteAll>;
+
+  /**
+   * Delete records from the index by either an array of ids, or a filter object.
+   * See [Filtering with metadata](https://docs.pinecone.io/docs/metadata-filtering#deleting-vectors-by-metadata-filter) for more on deleting records with filters.
+   *
+   * @example
+   * ```js
+   * const pinecone = new Pinecone();
+   * await pinecone.index('my-index').deleteMany(['record-1', 'record-2']);
+   *
+   * // or
+   * await pinecone.index('my-index').deleteMany({ filter: { genre: 'classical' }});
+   * ```
+   * @param options - An array of record id values or a filter object.
+   * @returns A promise that resolves when the delete is completed.
+   */
   deleteMany: ReturnType<typeof deleteMany>;
+
+  /**
+   * Delete a record from the index by id.
+   *
+   * @example
+   * ```js
+   * const pinecone = new Pinecone();
+   * await pinecone.index('my-index').deleteOne('record-1');
+   * ```
+   * @param options - The record is for deletion.
+   * @returns A promise that resolves when the delete is completed.
+   */
   deleteOne: ReturnType<typeof deleteOne>;
+
+  /**
+   * Describes the index's statistics such as total number of records, records per namespace, and the index's dimension size.
+   *
+   * @example
+   * ```js
+   * const pinecone = new Pinecone();
+   * await pinecone.index('my-index').describeIndexStats();
+   *
+   * // {
+   * //  namespaces: {
+   * //    '': { recordCount: 10 }
+   * //    foo: { recordCount: 2000 },
+   * //    bar: { recordCount: 2000 }
+   * //   },
+   * //   dimension: 1536,
+   * //   indexFullness: 0,
+   * //   totalRecordCount: 4010
+   * // }
+   * ```
+   * @returns A promise that resolve with the {@link IndexStatsDescription} when the operation is completed.
+   */
   describeIndexStats: ReturnType<typeof describeIndexStats>;
 
   private _fetchCommand: FetchCommand<T>;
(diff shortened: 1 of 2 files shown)


src/data/index.ts
```diff
@@ -89,22 +166,91 @@ export class Index<T extends RecordMetadata = RecordMetadata> {
...
+   * },{
+   *  id: 'record-2',
+   *  values: [0.176, 0.345, 0.263],
+   * }])
+   * ```
+   *
+   * @param data - An array of {@link PineconeRecord} objects to upsert.
+   * @returns A promise that resolves when the upsert is completed.
+   * @throws {@link Errors.PineconeConnectionError} when invalid environment, project id, or index name is configured.
+   * @throws {@link Errors.PineconeArgumentError} when invalid arguments are passed.
```
austin-denoble, line 202: I'm unsure whether we want to be annotating the different `@throws`, but I went ahead and added them for now.

jhamon: Seems useful. I wish typescript supported `throws` natively.

src/data/index.ts
```diff
@@ -48,23 +48,100 @@ export type {
...
+ */
 export class Index<T extends RecordMetadata = RecordMetadata> {
   private config: PineconeConfiguration;
   private target: {
     index: string;
     namespace: string;
   };
 
+  /**
+   * Delete all records from the index.
```
jhamon, line 62:
```suggestion
   * Delete all records from the targeted namespace. To delete all records across all namespaces, delete the index using {@link Pinecone.deleteIndex} and create a new one using {@link Pinecone.createIndex}.
   * 
   *
```

src/data/index.ts
```diff
@@ -48,23 +48,100 @@ export type {
...
+   *
+   * @example
```
jhamon, line 64: I think the example should demonstrate that only the targeted namespace is affected.

```typescript
const pinecone = new Pinecone()
await client.index('my-index').describeIndexStats()
// {
//   namespaces: { 
//     '': { recordCount: 9 }, 
//     foo: { recordCount: 1 } 
//   },
//   dimension: 8,
//   indexFullness: 0,
//   totalRecordCount: 10
// }

await client.index('my-index').deleteAll()

// Records in default namespace '' are now gone, but
// records in namespace 'foo' are not modified.
await client.index('my-index').describeIndexStats()
// {
//   namespaces: { foo: { recordCount: 1 } },
//   dimension: 8,
//   indexFullness: 0,
//   totalRecordCount: 1
// }
```

src/data/index.ts
```diff
@@ -48,23 +48,100 @@ export type {
...
+   * Delete records from the index by either an array of ids, or a filter object.
+   * See [Filtering with metadata](https://docs.pinecone.io/docs/metadata-filtering#deleting-vectors-by-metadata-filter) for more on deleting records with filters.
+   *
+   * @example
+   * ```js
+   * const pinecone = new Pinecone();
+   * await pinecone.index('my-index').deleteMany(['record-1', 'record-2']);
+   *
+   * // or
+   * await pinecone.index('my-index').deleteMany({ filter: { genre: 'classical' }});
```
jhamon, line 83:
```suggestion
   * await pinecone.index('my-index').deleteMany({ genre: 'classical' });
```

I'm pretty sure the `filter` key is not needed here. [It gets added before sending.](https://github.com/pinecone-io/pinecone-ts-client/blob/main/src/data/deleteMany.ts#L41)

src/data/index.ts
```diff
@@ -48,23 +48,100 @@ export type {
...
+
+  /**
+   * Delete a record from the index by id.
+   *
+   * @example
+   * ```js
+   * const pinecone = new Pinecone();
+   * await pinecone.index('my-index').deleteOne('record-1');
+   * ```
+   * @param options - The record is for deletion.
```
jhamon, line 98:
```suggestion
   * @param id - The id of the record to delete.
```

src/data/index.ts
```diff
@@ -48,23 +48,100 @@ export type {
...
+   * Instantiation of Index is handled by {@link Pinecone}
+   *
+   * @example
+   * ```js
+   * const pinecone = new Pinecone();
+   * const index = pinecone.index('my-index');
+   * ```
+   *
+   * @constructor
+   * @param indexName - The name of the index being configured.
```
jhamon, line 141:
```suggestion
   * @param indexName - The name of the index that will receive operations from this {@link Index} instance.
```

src/data/index.ts
```diff
@@ -89,22 +166,91 @@ export class Index<T extends RecordMetadata = RecordMetadata> {
...
+   * @returns A promise that resolves with the {@link FetchResponse} when the fetch is completed.
+   * @throws {@link Errors.PineconeConnectionError} when invalid environment, project id, or index name is configured.
+   * @throws {@link Errors.PineconeArgumentError} when invalid arguments are passed.
+   */
   async fetch(options: FetchOptions) {
     return await this._fetchCommand.run(options);
   }
 
+  /**
+   * Query records from the index by id or vector values.
```
jhamon, line 226:
```suggestion
   * Query records from the index. Query is used to find the `topK` records in the index whose vector values are most similar to the vector values of the query according to the distance metric you have configured for your index. See [Querying](https://docs.pinecone.io/docs/query-data)
```

jhamon (approved):
Thanks for filling these in. It's exciting seeing it come together. :)

jhamon (approved):
Thanks for filling these in. It's exciting seeing it come together. :)

src/data/index.ts
austin-denoble: Good catch.

src/data/index.ts
austin-denoble: Agreed, I think it's definitely helpful to have more examples involving namespace to make things more explicit.

Issue #117, opened 2023-09-14
[Feature] Utility Functions for Pinecone 1.0 


glody007:
### Is this your first time submitting a feature request?

- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing functionality

### Describe the feature

I'd like to address a specific concern regarding our project's compatibility with Pinecone 1.0. In the latest Pinecone release, functions such as chunkedUpsert and createIndexIfNotExists have been deprecated but are retained for retro-compatibility.

My question is simple: Are there plans to implement updated utility functions to replace these deprecated ones within our project, or is it expected that users will need to implement these changes independently to ensure Pinecone 1.0 compatibility?

Your insights on this matter would be greatly appreciated.

Thank you,

### Describe alternatives you've considered

These functions have been deprecated in Pinecone 1.0, which prompted me to take the initiative to reimplement them as utilities that are compatible with Pinecone 1.0.

### Who will this benefit?

It's worth noting that implementing updated utility functions to replace these deprecated ones could have significant benefits. Not only are these utilities very simple to use, but they also have the potential to expedite the development process for all users, enhancing productivity and ensuring a smooth transition to Pinecone 1.0.

### Are you interested in contributing this feature?

I'm more than willing to contribute to the integration process and help with any necessary modifications.

jhamon:
Thanks for the question. We're definitely interested in making the experience as smooth as possible.

My general feeling about the legacy utils is that although they are useful, especially when repeatedly doing those actions in integration testing (which is what they were originally used for), the fact they were/are needed reflects shortcomings in the client methods itself. I think ideally you should have the client do whatever it is you need without resorting to additional utility wrappers.

## waitUntilIndexIsReady

This utility handled polling for status to see if the index is ready to handle data operations. This should be covered with the newly added `waitUntilReady` boolean option accepted by `createIndex`.

```typescript
import { Pinecone } from '@pinecone-database/pinecone'

const pinecone = new Pinecone()
await pinecone.createIndex({ 
  name: 'my-index', 
  dimension: 1586, 
  waitUntilReady: true  // yay, this exists today
})
```

## createIndexIfNotExists

The legacy utility implementation makes 2 API calls:
- Calls to get the list of indexes
- Calls to create the index if missing from list

[...]

PR #125, opened 2023-09-21
Expose an option in createIndex to suppress the 409 conflict error

glody007:
## Problem
To create an index if it does not exist requires two API calls:

- Calls to get the list of indexes
- Calls to create the index if missing from the list

## Solution
Instead of making two separate API calls, the idea is to streamline the process into a single call by introducing an option within the `createIndex` function.

Specifically, this solution proposes the addition of an option called `suppressConflicts`  that allows the function to suppress the "409 Conflict" error. By using this option, the createIndex function would attempt to create the index, but if it encounters a "409 Conflict" error (indicating that the index is already present), it would not raise an exception This would reduce the need for additional calls or error handling in the client code.

## Type of change

## Test Plan
Adding a test case to check if a specific error for conflict is raised when the `suppressConflicts` option is not `true`


jhamon:
Thanks for the quick follow up to #117 

This code looks good to me. The CI errors are just because PRs from forks can't access secrets (e.g. our Pinecone API credentials) needed to run integration tests so you don't need to worry about those. 

The only other thing I'd like to see added to this is a unit test covering this 409 error case added in [here](https://github.com/pinecone-io/pinecone-ts-client/blob/main/src/control/__tests__/createIndex.test.ts#L153).

glody007:
Thank you for your feedback. I've already added a unit test to cover the 409 error case, and I've provided additional details in the PR description to make it more comprehensive. Please feel free to review the changes, and if there are any further suggestions or adjustments needed, I'm ready to make them. Your guidance is invaluable in helping me improve my code. I truly appreciate your input and suggestions.

Issue #124, opened 2023-09-19
[Bug] v1 Index<T>.query throwing TypeError: Cannot read properties of undefined (reading 'text') during migration


mattgraphlan:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

I'm migrating to v1 as described in https://github.com/pinecone-io/pinecone-ts-client/blob/main/v1-migration.md#query

I can upsert after migrating just fine, but I can't query.  My code is as follows:

```ts
    const namespace = ""
    const topK = 10
    const queryRequest = {
      vector: embeddings,
      topK,
      includeMetadata: true,
      includeValues: true,
    } as QueryByVectorValues;
    try {
      // Query the index with the defined request
      const ns = index.namespace(namespace);
      const queryResult = await ns.query(queryRequest);
      return queryResult.matches || [];
    } catch (e) {
      console.error("Error querying embeddings: ", e, (e as Error).stack);
      throw new Error(`Error querying embeddings: ${e}`);
    }
```

I have verified that there are actually embeddings (`RecordValues`).

The stack trace is a bit unhelpful: 
```
TypeError: Cannot read properties of undefined (reading 'text')
    at eval (webpack-internal:///(rsc)/./node_modules/@pinecone-database/pinecone/dist/errors/utils.js:136:40)
    at step (webpack-internal:///(rsc)/./node_modules/@pinecone-database/pinecone/dist/errors/utils.js:107:23)
    at Object.eval [as next] (webpack-internal:///(rsc)/./node_modules/@pinecone-database/pinecone/dist/errors/utils.js:48:20)
    at eval (webpack-internal:///(rsc)/./node_modules/@pinecone-database/pinecone/dist/errors/utils.js:26:71)
    at new Promise (<anonymous>)
    at __awaiter (webpack-internal:///(rsc)/./node_modules/@pinecone-database/pinecone/dist/errors/utils.js:8:12)
    at extractMessage (webpack-internal:///(rsc)/./node_modules/@pinecone-database/pinecone/dist/errors/utils.js:129:12)
    at eval (webpack-internal:///(rsc)/./node_modules/@pinecone-database/pinecone/dist/errors/handling.js:178:76)
    at step (webpack-internal:///(rsc)/./node_modules/@pinecone-database/pinecone/dist/errors/handling.js:107:23)
[...]

PR #126, opened 2023-09-26
[Bugfix] Implement next.js fetch compatibility

jhamon:
## Problem

Next.js seems to stub out the cross-fetch polyfill which has the effect of silently breaking code which depends on cross-fetch to make API requests.

Related bugs: 
- #124 
- v1 migration problems in https://github.com/pinecone-io/pinecone-vercel-starter/pull/14

## Solution

I have added a `getFetch` utility which looks for a `fetch` implementation already loaded in the global scope and returns the cross-fetch implementation only as a last resort. 

- Now when using our client library in a next.js project, this should result in the `@vercel/fetch` implementation being used in favor of the disabled polyfill. 
- When using this client library in other environments where next.js is not being used, it still activates the cross-fetch polyfill.
- Also, I've exposed an experimental configuration option for anyone who wants to pass in a different fetch implementation to use. This could open up some flexibility for people with specific network configurations who need the ability to configure the http client being used in some way. If the user provides a fetch implementation through the `fetchApi` config option, it will take priority over a fetch in the global scope or the cross-fetch polyfill.

### Other changes

- Added additional integration tests for `query` method
- Added an explicit dependency on the `encoding` module to resolve a warning message [similar to this one](https://github.com/vercel/next.js/discussions/54109)

## Type of Change

## Test Plan

I am currently testing this change in the context of work-in-progress to migrate our vercel-starter example to using the v1 version of the client. https://github.com/pinecone-io/pinecone-vercel-starter/pull/14

- Checkout the PR branch from pinecone-vercel-starter in a sibling directory
- Install my local changes to this client with `npm install ../pinecone-ts-client`
- Try to run the starter project with `npm run dev`
- Attempt to use the sample app and observe it progresses beyond the error related broken cross-fetch

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,7 +10,7 @@ module.exports = {
   testPathIgnorePatterns: ['src/integration'],
   testTimeout: 100000,
   verbose: true,
-  detectOpenHandles: true,
+  detectOpenHandles: false,
   collectCoverageFrom: [
     '<rootDir>/src/**/*.ts',
     '!**/src/pinecone-generated-ts-fetch/**',
--- a/src/integration/data/deleteMany.test.ts
+++ b/src/integration/data/deleteMany.test.ts
@@ -1,9 +1,5 @@
 import { Pinecone, Index } from '../../index';
-import {
-  randomString,
-  createIndexIfDoesNotExist,
-  generateRecords,
-} from '../test-helpers';
+import { randomString, generateRecords } from '../test-helpers';
 
 describe('deleteMany', () => {
   const INDEX_NAME = 'ts-integration';
@@ -12,7 +8,12 @@ describe('deleteMany', () => {
   beforeAll(async () => {
     pinecone = new Pinecone();
 
-    await createIndexIfDoesNotExist(pinecone, INDEX_NAME);
+    await pinecone.createIndex({
+      name: INDEX_NAME,
+      dimension: 5,
+      waitUntilReady: true,
+      suppressConflicts: true,
+    });
 
     namespace = randomString(16);
     index = pinecone.index(INDEX_NAME);
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -5,12 +5,12 @@ export const getFetch = (config: PineconeConfiguration) => {
   if (config.fetchApi) {
     // User-provided fetch implementation, if any, takes precedence.
     return config.fetchApi;
-  } else if (fetch) {
+  } else if (global.fetch) {
     // If a fetch implementation is already present in the global
     // scope, use that. This should prevent confusing failures in
     // nextjs projects where @vercel/fetch is mandated and
     // other implementations are stubbed out.
-    return fetch;
+    return global.fetch;
   } else {
     // Use polyfill as last resort
     return crossFetch;
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -1,0 +1,18 @@
+import crossFetch from 'cross-fetch';
+import type { PineconeConfiguration } from '../data';
+
+export const getFetch = (config: PineconeConfiguration) => {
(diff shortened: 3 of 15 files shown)


jest.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   testPathIgnorePatterns: ['src/integration'],
   testTimeout: 100000,
   verbose: true,
-  detectOpenHandles: true,
+  detectOpenHandles: false,
```
jhamon, line 13: I started getting an error related to TLSWRAP after adding some more integration tests. I should follow up on this later, but for now I don't want to block on it.

src/utils/fetch.ts
```diff
@@ -0,0 +1,18 @@
...
+    // User-provided fetch implementation, if any, takes precedence.
+    return config.fetchApi;
+  } else if (global.fetch) {
+    // If a fetch implementation is already present in the global
+    // scope, use that. This should prevent confusing failures in
+    // nextjs projects where @vercel/fetch is mandated and
+    // other implementations are stubbed out.
+    return global.fetch;
+  } else {
+    // Use ponyfill as last resort
```
zackproser, line 15:
```suggestion
    // Use polyfill as last resort
```

src/utils/fetch.ts
```diff
@@ -0,0 +1,18 @@
...
+    return crossFetch;
```
zackproser, line 16: So just to confirm my understanding: 
1. If the user provides fetch, use that
2. In Vercel, the global.fetch if statement will be true, so we're good in Vercel land (or do they actually provide `config.fetchApi`?)
3. When running locally in a Pinecone example app, `cross-fetch` will be used, so that network requests the client needs to initialize will succeed

Do I have that right?

src/utils/fetch.ts
```diff
@@ -0,0 +1,18 @@
+import crossFetch from 'cross-fetch';
+import type { PineconeConfiguration } from '../data';
+
+export const getFetch = (config: PineconeConfiguration) => {
```
austin-denoble, line 4: **question:** Is it worth adding a tiny unit test file for this utility?

austin-denoble (approved):
Nice fix, Jen. Overall the code LGTM, I did run into issues while trying to test via the Vercel starter directly though:

Screenshot 2023-09-27 at 11 12 42 AM
Screenshot 2023-09-27 at 11 12 32 AM

This may be an issue on my end, still playing with it a bit. Let me know if there's anything obvious.

src/integration/data/deleteMany.test.ts
```diff
@@ -12,7 +8,12 @@ describe('deleteMany', () => {
   beforeAll(async () => {
     pinecone = new Pinecone();
 
-    await createIndexIfDoesNotExist(pinecone, INDEX_NAME);
```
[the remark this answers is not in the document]
jhamon: Yep, sorry I didn't explain that better.

src/utils/fetch.ts
jhamon: My understanding is that next.js loads `@vercel/fetch` and it will be available as a global without the need to specifically import it. So the next.js scenario would be the second logical branch here.

The ability to pass it through config is just an escape hatch for anyone that has weird/specific needs around the http client configuration that I don't have bandwidth to address at the moment.

src/utils/fetch.ts
jhamon: This is simple enough that I don't think unit-testing would tell us much. I think what would be more valuable here would be to have some next.js-specific integration testing. But I think it's too much additional scope to add as part of this PR. Will have to look into it as a follow-up item.

src/utils/fetch.ts
zackproser: Roger, thanks!

PR #129, opened 2023-09-30
Refactoring for better typedoc output

jhamon:
## Problem

The main problem I initially set out to fix was inconsistency between some class methods being shown as properties in the rendered typedoc output. Of course, I once I started tinkering with docs, I wanted to change 50 other things.

## Solution

- Small refactor of `Pinecone` and `Index` classes to define "normal" methods instead of assigning function properties in the constructor. This should have no impact to behavior, but allows typedoc to properly and consistently categorize methods on these objects.
- Added a CSS file to start tweaking the appearance. It's dev-design 😬, but it's a start.
- Added a bunch of missing docstrings, migrate some info from README into docstrings
- Tweaked typedoc config to add some sidebar links and hide items marked `@internal` 
- Migrated `@param options.foo` information from the methods where these options params are passed (e.g. from `createIndex()`) onto the options type definition itself (e.g. `CreateIndexOptions`). I did this because docstring declarations like `@param options.foo` just get ignored by typedoc.
  
## Type of Change

### `Pinecone`

Before | After
-------|------
Screenshot 2023-09-30 at 3 49 18 PM | Screenshot 2023-09-30 at 3 48 49 PM

### `Index`

Before | After
-------|------
Screenshot 2023-09-30 at 3 43 26 PM | Screenshot 2023-09-30 at 3 43 12 PM


src/control/createIndex.ts
```diff
@@ -51,7 +82,7 @@ export const createIndex = (api: IndexOperationsApi) => {
     'createIndex'
   );
 
-  return async (options: CreateIndexOptions): Promise<string | void> => {
```
jhamon, line 54: This just seemed incorrect, so I fixed it while I was here 🤷

src/control/types.ts
```diff
@@ -30,3 +30,21 @@ export const MetadataConfigSchema = Type.Object(
...
+  | 's1.x8'
+  | 'p1.x1'
+  | 'p1.x2'
+  | 'p1.x4'
+  | 'p1.x8'
+  | 'p2.x1'
+  | 'p2.x2'
+  | 'p2.x4'
+  | 'p2.x8'
+  | string;
```
jhamon, line 50: This type includes `string` so isn't really a narrowing of the type in comparison to the past. But I thought it would make for better documentation to show that actual valid values here.

src/data/index.ts
```diff
@@ -126,7 +194,11 @@ export class Index<T extends RecordMetadata = RecordMetadata> {
    * @returns A promise that resolves when the delete is completed.
    * @throws {@link Errors.PineconeArgumentError} when invalid arguments are passed.
    */
-  deleteOne: ReturnType<typeof deleteOne>;
+  deleteOne(id) {
+    return this._deleteOne(id);
+  }
+  /** @hidden */
+  _deleteOne: ReturnType<typeof deleteOne>;
```
jhamon, line 201: This and all similar changes to `Index` and `Pinecone` are about helping typedoc display method information in a consistent way. If you assign a function to a property in the constructor, it treats it as a property instead of as a method and having inconsistency there made the docs look ugly/confusing.

src/pinecone.ts
```diff
@@ -157,35 +216,39 @@ export class Pinecone {
...
    * The minimum required configuration to create an index is the index name and dimension.
    * ```js
-   * await client.createIndex({ name: 'my-index', dimension: 128 })
+   * await pinecone.createIndex({ name: 'my-index', dimension: 128 })
    * ```
    * @example
    * In a more expansive example, you can specify the metric, number of pods, number of replicas, and pod type.
    * ```js
-   * await client.createIndex({
+   * await pinecone.createIndex({
```
jhamon, line 249: As a style thing, I'm trying to always reference `Pinecone` instances with a variable called `pinecone`.

src/pinecone.ts
```diff
@@ -157,35 +216,39 @@ export class Pinecone {
...
    *  name: 'my-index',
-   *  dimension: 128,
+   *  dimension: 1536,
```
jhamon, line 251: This is the output dimension of OpenAI embedding models, so is more "realistic" than other arbitrary numbers we might pick.

src/index.ts
```diff
@@ -47,6 +48,12 @@ export type {
   ScoredPineconeRecord,
 } from './data';
 export type { FetchAPI } from './pinecone-generated-ts-fetch';
+export type {
+  IndexMeta,
+  IndexMetaStatus,
+  IndexMetaStatusStateEnum,
+  IndexMetaDatabase,
+} from './pinecone-generated-ts-fetch';
```
jhamon, line 56: The docs build was complaining about these being missing. So I figured I should export them.

src/data/fetch.ts
```diff
@@ -6,11 +6,19 @@ import type { PineconeRecord, RecordId, RecordMetadata } from './types';
...
+/** The list of record ids you would like to fetch using { @link Index.fetch } */
 export type FetchOptions = Array<RecordId>;
 
+/** The response from {@link Index.fetch } */
 export type FetchResponse<T extends RecordMetadata = RecordMetadata> = {
-  records?: { [key: string]: PineconeRecord<T> };
-  namespace?: string;
+  /** A map of fetched records, keyed by record id. */
+  records: {
+    [key: RecordId]: PineconeRecord<T>;
```
austin-denoble, line 17: It looks like TS `4.1`, `4.2`, and `4.3` don't support using a type alias as an index signature parameters type. 🤔 

Sort of annoying:
Screenshot 2023-10-02 at 11 03 28 AM

jhamon: Annoying, but at least it's an easy fix to put it back how it was.

jhamon: Also: hooray for tests catching problems.

PR #132, opened 2023-10-03
Add favicon to TypeDoc generated output via plugin script

austin-denoble:
## Problem
Currently, our generated documentation for the Typescript client doesn't have a favicon, which would be nice. TypeDoc doesn't provide a straightforward way of providing a favicon asset, but they do provide the ability to write custom plugin scripts to modify the generated output.

TypeDoc custom theme documentation: https://github.com/TypeStrong/typedoc/blob/master/internal-docs/custom-themes.md

## Solution

- Add an `/assets/` folder to the top level of the repo.
- Add the `favicon-32x32.png` image (generated from: https://www.figma.com/file/4tZhnN7BbFM6yWqPGjjVbf/Favicon?type=design&node-id=0-1&mode=design&t=nNcepCxWYRcJWwWx-0)
- Add a new `docs-theme.mjs` script to be provided to TypeDoc at build time.

The `docs-theme.mjs` script does two things:
- `onPageRendered` takes the `page` output from typedoc and injects a `<link .../>` into `head`.
- Copies `favicon-32x32.png` from `./assets` to the new `/docs` folder.

**Note:** I was having issues writing a plain `.js` file for the script (no imports warning, etc), but `.mjs` seems to work well given our current TS setup. Would definitely love feedback here around improving things though, I didn't play around with our configs much.

|Before|After|
|-----|-----|
|Screenshot 2023-10-03 at 5 45 39 PM|Screenshot 2023-10-03 at 5 45 44 PM|

## Type of Change

## Test Plan
You should be able to test locally by pulling this branch down, and running `npm run docs:build`. Make sure the generated output includes the Pinecone logo in the tab. Inspect the page source to confirm the presence of `favicon-32x32.png`.

Confirm the docs build step of the PR workflow run completes as expected.


.eslintrc.json
```diff
@@ -1,7 +1,8 @@
 {
   "env": {
     "browser": true,
-    "es2021": true
+    "es2021": true,
+    "node": true
```
austin-denoble, line 5: I needed to add `node` here to get lint to stop whining about `process` not being accessible in `docs-theme.mjs`.

assets/docs-theme.mjs
```diff
@@ -0,0 +1,38 @@
+import { copyFileSync } from 'fs';
```
austin-denoble, line 1: It's been a minute since I've had to deal with commonjs vs. modules vs. ts config. I'm not sure if an `.mjs` file was the best way to do this, and I also wasn't sure if there were specific considerations around adding a script that's outside `/src/` and used in CI / build. Definitely open to changing things up here.

jhamon (approved):
Looks great 🌲

PR #133, opened 2023-10-06
Fix typedoc Light/Dark mode styling

austin-denoble:
## Problem
@tdonia noticed the dark mode for the typescript client documentation was broken as we overrode a variety of styles with custom css, but didn't account for how the typedoc default theme handles media queries. We had also hidden the light/dark mode toggle, and I had cookies cached locally setting things to light mode, so this went unnoticed for a bit. 😅 

There's no easy way to disable the default light/dark toggle behavior, and it defaults to the user's `OS` value.

## Solution
- Re-expose the mode selector. We had also hidden the entire settings accordion.
- Update `docs-styles.css` to handle media queries on `prefers-color-scheme` and the `data-theme` attribute.

Initially I wanted to override everything with our own colors and keep the light/dark toggle disabled, but I think this would be far more tedious than just copying their patterns for managing colors, and overriding things individually for each mode.

In this PR I'm basically inheriting all their dark mode colors, and applying our Pinecone stuff in light mode. I also tweaked a few things to clean up look and feel in a few places while I was in there.

|Before|After|
|-----|-----|
|Screenshot 2023-10-06 at 3 41 49 PM|Screenshot 2023-10-06 at 3 42 09 PM|

## Type of Change

## Test Plan
You can pull the branch down and run `npm run docs:build` locally. Navigate to `./docs` in the repo and open the index file to view the generated documentation. Validate that light and dark modes both look visually reasonable.

### DIFF
--- a/docs-styles.css
+++ b/docs-styles.css
@@ -1,26 +1,88 @@
+/* 
+  The media query pattern here is mimicking what typedoc does within their default theme.
+  Currently you cannot disable the dark / light mode toggle easily, so this provides a
+  way of applying specific overrides in the context of their default styling.
+  
+  https://github.com/TypeStrong/typedoc/blob/master/static/style.css
+*/
 :root {
-  /* I copied these vars from docs.pinecone.io */
+  /* Styles copied from docs.pinecone.io */
   --font-family: 'MediumLLWeb', sans-serif;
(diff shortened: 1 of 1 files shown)


tdonia:
looks great // works for me in both modes.  thanks @austin-denoble!

jhamon (approved):
Thanks for tidying up :shipit:

PR #135, opened 2023-10-06
Overhaul error handling to address text() error

jhamon:
## Problem

Some users are still running into errors that can't be properly diagnosed because of a bug in error handling causing the real error to get obscured by an unrelated problem. There's a point at which we assume we should be dealing with a `ResponseError` with a `text()` field; in reality, that's not a correct assumption so it's not safe to cast to that type.

- #124 

## Solution

This was one of those changes where you pull on the thread and the whole carpet unravels.

- Refactor error handling away from nested callbacks into a flat error handler function
- Wrap the underlying exception by attaching it as a `cause` whenever a `PineconeConnectionError` exception is being thrown. This will help us investigate #124 
- Migrate away from method-specific error handling to a consistent approach based on middleware functions. We already had an httpErrorMapper, but with this approach it only has to get wired up twice instead of like 20 times so that really simplifies what we need to cover in unit testing.
- Implement a bunch of new integration tests to give confidence the middleware is actually working as expected.
    - The cost of this change is we lose a little bit of convenience in the 404 case where we were fetching and presenting the list of valid indexes/collections. But that small bit of convenience doesn't seem worth a radically more complex approach to error handling.
    - We no longer need to test exception handling in most method implementations
    - It unlocks the ability for some really cool debug options. Try `PINECONE_DEBUG=true` and `PINECONE_DEBUG_CURL=true` 

### `PINECONE_DEBUG=true` and `PINECONE_DEBUG_CURL=true` 

Screenshot 2023-10-06 at 5 32 44 PM

## Type of Change

## Test Plan


src/errors/request.ts
```diff
@@ -8,13 +9,56 @@ import { BasePineconeError } from './base';
  * - Incorrect configuration of the client. The client builds its connection url using values supplied in configuration, so if these values are incorrect the request will not reach a Pinecone server.
  * - An outage of Pinecone's APIs. See [Pinecone's status page](https://status.pinecone.io/) to find out whether there is an ongoing incident.
  *
+ * The `cause` property will contain a reference to the underlying error. Inspect its value to find out more about the root cause of the error.
```
austin-denoble, line 12: **praise:** Thank you for the docstring date. 👍

src/utils/middleware.ts
```diff
@@ -0,0 +1,24 @@
+import { ResponseError } from '../pinecone-generated-ts-fetch';
+import { handleApiError } from '../errors';
+
+export const middleware = [
+  {
+    onError: async (context) => {
```
austin-denoble, line 6: **praise:** Huge praise for all this cleanup, I feel like errors are much more straightforward to reason around now, and it's really helpful to pull all the testing into a more centralized place rather than having it spread across all the individual methods.

src/utils/testHelper.ts
```diff
@@ -0,0 +1,11 @@
+import { ResponseError } from '../pinecone-generated-ts-fetch';
```
austin-denoble, line 1: **question:** Can this file be removed? Doesn't seem like it's actually being used.

src/integration/control/configureIndex.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      dimension: 5,
+      waitUntilReady: true,
+      podType: 'p1.x1',
+      replicas: 2,
+    });
+  });
+
+  afterEach(async () => {
+    await pinecone.deleteIndex(indexName);
+  });
```
austin-denoble, line 24: **question:** Do we need to be creating and deleting an index for each test rather than using `beforeAll` and `afterAll`? It seems like this would lead to slower tests as we have to wait somewhere around 10 seconds for each index to spin up for each test. For this specific suite could we not just create an index once, perform all the tests and operations, and then delete `afterAll` for cleanup?

src/integration/errorHandling.test.ts
```diff
@@ -0,0 +1,82 @@
...
+        expect(err.message).toEqual(
+          'Request failed to reach Pinecone. This can occur for reasons such as incorrect configuration (environment, project id, index name), network problems that prevent the request from being completed, or a Pinecone API outage. Check your client configuration, check your network connection, and visit https://status.pinecone.io/ to see whether any outages are ongoing.'
+        );
+        expect(err.cause).toBeDefined();
+      }
+    });
+
+    test('calling data plane', async () => {
+      const p = new Pinecone({
+        apiKey: 'api-key-2',
```
austin-denoble, line 26: **question:** Should this still be `process.env.PINECONE_API_KEY`? Asking as the section is described as `when environment is wrong`.

austin-denoble (approved):
Fantastic work, really appreciate this cleanup. I found it's much easier overall to reason about how errors are flowing through things now as opposed to before. I also really appreciate adding in additional integration tests while you're here.  

Just a few questions, but overall looks good to me. Having error handling and throwing centralized rather than all over the place feels good. 🚢

src/utils/testHelper.ts
jhamon: Yeah, I think it can be removed. I had created a new test-helper in here but later I ended up deleting all the unit tests that were using it. 😂

src/integration/control/configureIndex.test.ts
jhamon: As currently written, this test suite does end up adding quite a bit of time due to the index creation. But for now I want to keep every test fully isolated from others. 

There seems to be some timing issues in this configure endpoint when called multiple times at a row. At one point I had a test here that scaled replicas up and then down again (using the same index) and it was throwing a 500 error at a high rate, even when I waited for status to reach the Ready state. I don't think they envisioned people scaling up and down separated only by milliseconds.

src/integration/errorHandling.test.ts
jhamon: I can change it to make it more clear that the environment is the only thing that matters here. Ultimately only the `environment` value matters for this test since it's an input into the url and there is no `wrong-environment` environment in production.

PR #136, opened 2023-10-06
Add `PINECONE_DEBUG` request middleware

jhamon:
## Problem

Sometimes it is helpful to see exactly what requests are being made under the hood.

## Solution

Define new middleware functions that log out helpful information.

Set new env vars for extra output:

```sh
PINECONE_DEBUG=true
PINECONE_DEBUG_CURL=true
```

Screenshot 2023-10-06 at 5 32 44 PM

## Type of Change

- [x] None of the above: developer-oriented feature

## Test Plan


src/utils/middleware.ts
```diff
@@ -1,7 +1,78 @@
...
+}
+
+/**
+ * Enable the `PINECONE_DEBUG_CURL` environment variable to print the equivalent
+ * curl commands for each request. These commands will include the API key and
+ * other sensitive information, so be careful when using this option.
+ */
+if (process && process.env && process.env.PINECONE_DEBUG_CURL) {
+  const debugCurlMiddleware = {
+    post: async (context) => {
```
austin-denoble, line 59: **question:** I'm assuming the way it works on the OpenAPI side is if there are multiple `post` elements they're just executed sequentially on the specified event.

Was there documentation or anything you found for OpenAPI and how they support middleware, or was it all code spelunking? I'd be curious to take a look myself at some point.

austin-denoble (approved):
This is great! Will be extremely helpful for testing and debugging overall. :shipit:

src/utils/middleware.ts
jhamon: Just code spelunking. Relevant code is all in this [fetch method](https://github.com/pinecone-io/pinecone-ts-client/blob/main/src/pinecone-generated-ts-fetch/runtime.ts#L194). 

They are expecting `config.middleware` to be an array of middleware objects with optional `pre`, `onError`, and `post` callbacks. All are invoked, with the only caveat is if one of the middlewares throws an exception then nothing else gets called. This means you can only really have a single `onError` specified.

PR #144, opened 2023-10-20
Allow gh-pages deployment to persist CNAME file

austin-denoble:
## Problem
We've been looking at adding custom domains to our documentation references which are hosted via GitHub pages. When applying a custom domain, GitHub modifies the source branch and commits a CNAME file automatically.

See here: https://github.com/pinecone-io/pinecone-ts-client/blob/gh-pages/CNAME

If you're deploying to this branch as a part of CI, you need to handle maintaining the CNAME file. We're currently using the `git-publish-subdir-action` to achieve this: https://github.com/marketplace/actions/push-git-subdirectory-as-branch.  This action blows away everything unless you specify specific files to persist.

## Solution
The action we're using supports passing a `CLEAR_GLOBS_FILE` argument allowing you to control what's replaced and what's persisted: https://github.com/marketplace/actions/push-git-subdirectory-as-branch#custom-clear-operations

- Add `.clear-target-files` under `.github/`. This file tells the action to clear everything but persist the CNAME file.
- Update `merge.yml` to pass the file to the action in the merge workflow -> docs deployment step.

## Type of Change

## Test Plan
Once this change is merged we can check the `gh-pages` branch to make sure the CNAME file is not blown away when we regenerate things and push to `gh-pages`.


.github/.clear-target-files
```diff
@@ -0,0 +1,2 @@
+**/*
```
austin-denoble, line 1: I figured putting this under `.github` since it's specifically CI related was fine.

PR #147, opened 2023-10-23
Check for `process` existence before accessing it

omarestrella:
## Problem

On some edge platforms (specifically Cloudflare Workers), `process` does not exist at all, so checking `!process` isn't enough and causes a runtime error / crash.

## Solution

I added a `typeof process` check before any `process` access occurs so that the Pinecone client works on non-Node runtimes.

## Type of Change

## Test Plan

No test changes should be necessary, as this is just a more strict `!process` check.

### DIFF
--- a/src/pinecone.ts
+++ b/src/pinecone.ts
@@ -152,7 +152,7 @@ export class Pinecone {
    * @returns A {@link PineconeConfiguration} object populated with values found in environment variables.
    */
   _readEnvironmentConfig(): PineconeConfiguration {
-    if (!process || !process.env) {
+    if (typeof process === 'undefined' || process || !process.env) {
       throw new PineconeEnvironmentVarsNotSupportedError(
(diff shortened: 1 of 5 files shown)


jhamon:
Thanks for the fix. If you run `npm run format` and commit the diff that will fix the lint/format check. The rest are failing due to environment variables being withheld for security reasons so shouldn't be a concern.

omarestrella:
> Thanks for the fix. If you run `npm run format` and commit the diff that will fix the lint/format check. The rest are failing due to environment variables being withheld for security reasons so shouldn't be a concern.

Gotcha, formatting should be fixed.

austin-denoble:
This fix has been included in the v1.1.2 release: https://github.com/pinecone-io/pinecone-ts-client/releases/tag/v1.1.2. 

Thanks again for contributing!

PR #148, opened 2023-10-25
Refactor projectIdSingleton -> hostUrlSingleton

austin-denoble:
## Problem
In the future, we'll be using `describeIndex` to retrieve the specific `host` value for a given Index. Previously, this was handled internally by the client via configuration values and pinging `whoami`, and caching the results to prevent unneeded requests and latency.

We want to maintain this caching behavior for `apiKey` -> `indexName` while simplifying the client configuration and calling `describeIndex` to retrieve the `host`.

## Solution
The new `IndexHostSingleton` replaces the existing `ProjectIdSingleton`. The functionality is very similar between the two singletons at a high level, but the code paths for making network calls are different. I've also gone ahead and pulled out `projectId` and `environment` as configuration values as they'll no longer be needed to generate the data plane address.

- Delete `ProjectIdSingleton` and add new `IndexHostSingleton`.
- Add unit tests for `IndexHostSingleton`.
- Remove `projectId` and `environment` from `PineconeConfiguration` and update related areas, tests, documentation bits, etc. Add new `PINECONE_CONTROLLER_HOST` environment variable.
- Add `.then()` callbacks to `describeIndex` and `deleteIndex` to both set and delete the host from the cache when needed.

## Type of Change

## Test Plan
Testing is a bit tricky at the moment, but you should be able to pull the branch down and test using the repl:

```
npm run repl

await init()
await client.listIndexes()

await client.index('my-index').describeIndexStats()
```

Make sure unit tests are passing in CI.

Integration tests are a bit borked right now, but there is a follow-up ticket to address those specifically.


src/__tests__/pinecone.test.ts
```diff
@@ -14,48 +16,38 @@ jest.mock('../utils', () => {
 });
 jest.mock('../data/fetch');
 jest.mock('../data/upsert');
+jest.mock('../data/hostUrlSingleton');
+
+jest.mock('../control', () => {
```
austin-denoble, line 21: I spent a lot of time trying to get the mocking for things working and it still feels like a mess. 😭

src/integration/errorHandling.test.ts
```diff
@@ -6,7 +6,6 @@ describe('Error handling', () => {
     test('calling control plane', async () => {
```
austin-denoble, line 6: This file will need to be fixed up.

src/integration/initialization.test.ts
```diff
@@ -4,7 +4,6 @@ describe('Client initialization', () => {
   test('can accept a config object', () => {
```
austin-denoble, line 4: Same as above, just fixing up lint / typescript for now.

src/control/describeIndex.ts
```diff
@@ -9,7 +9,13 @@ export type DescribeIndexOptions = IndexName;
 /** The description of your index returned from { @link Pinecone.describeIndex } */
 export type IndexDescription = IndexMeta;
 
-export const describeIndex = (api: IndexOperationsApi) => {
+export const describeIndex = (
+  api: IndexOperationsApi,
+  callback?: (
```
austin-denoble, line 14: I went back and forth on a few different means of allowing the `describeIndex` call to interact with `HostUrlSingleton`. Initially I wanted to use it within here, but then we didn't have access to the `PineconeConfiguration` for the `cacheKey` in set, so went with a callback.

src/data/hostUrlSingleton.ts
```diff
@@ -0,0 +1,89 @@
...
+    indexName: string
+  ): Promise<string> => {
+    if (!indexOperationsApi) {
+      indexOperationsApi = _buildIndexOperationsApi(config);
+    }
+
+    const describeResponse = await describeIndex(indexOperationsApi)(indexName);
+    const host = describeResponse.status?.host;
+
+    if (!host) {
```
jhamon, line 47: It's fine to cover this case here because the type indicates it could occur (mostly because of loose typing in the openapi spec, I think), but in practice I have never seen the host value come back missing. I think the url is known immediately, whether or not the index is actually ready to receive data plane calls.

src/data/hostUrlSingleton.ts
```diff
@@ -0,0 +1,89 @@
...
+// We use describeIndex to retrieve the data plane url (host) for a given API key
+// and index. We only ever want to call describeIndex a maximum of once per API key
+// and index so we cache them in a singleton for reuse.
+export const HostUrlSingleton = (function () {
+  const hostUrls = {}; // map of apiKey-indexName to hostUrl
+  let indexOperationsApi: InstanceType<typeof IndexOperationsApi> | null = null;
+
+  const _buildIndexOperationsApi = (config: PineconeConfiguration) => {
+    const { apiKey } = config;
+    const controllerPath = `https://api.pinecone.io`;
```
jhamon, line 21: Some way of overriding this value using an env var like `PINECONE_CONTROLLER_HOST` would be useful in getting the integration tests working over the short-term while api.pinecone.io is still being worked on.

jhamon (approved):
Looking over this diff really made me appreciate what a trooper you've been doing so many huge reviews for me lately.

To summarize my feedback:
- Add an env var to override control plane url
- Consider extracting a new class of some kind to handle config/initialization for IndexOperationsApi
- You can simplify a lot of stuff by leaving describeIndex as is and making it the responsibility of the Pinecone class to inform the singleton of the host whenever pinecone.describeIndex() gets called
- There's an edge case with pinecone.deleteIndex also. Probably things should be evicted from the singleton cache when an index is deleted. Otherwise repeated create/delete/create operations will leave you in a state where the cached host url is invalid.

src/pinecone.ts
```diff
@@ -134,12 +134,23 @@ export class Pinecone {
...
+   * @internal
+   * Used as a callback by {@link Pinecone.describeIndex} to cache any index host URLs that are returned.
+   * This bypasses the need to fetch the host on subsequent dataplane calls.
+   */
+  _handleDescribeIndex = (response: IndexDescription, indexName: IndexName) => {
+    if (response.status?.host) {
+      HostUrlSingleton._set(this.config, indexName, response.status.host);
+    }
+  };
+
```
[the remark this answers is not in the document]
austin-denoble: Good point, not really sure why I decided to go with a callback at this point. This would have been much cleaner. 😂

src/control/indexOperationsBuilder.ts
```diff
@@ -0,0 +1,27 @@
+import {
```
austin-denoble, line 1: This seemed a bit silly to unit test, and the only thing we'd really care about checking is the proper `controllerPath` being set. Since the internals of the OpenAPI classes are protected, that's not necessarily easy unless we want to use a `ts-ignore`.

src/pinecone.ts
```diff
@@ -219,7 +200,17 @@ export class Pinecone {
    * @returns A promise that resolves to {@link IndexMeta}
    */
   describeIndex(indexName: IndexName) {
-    return this._describeIndex(indexName);
+    const describeIndexPromise = this._describeIndex(indexName);
```
austin-denoble, line 203: Swapped from passing a callback to appending a `.then()` on the promise before returning.

src/pinecone.ts
```diff
@@ -320,7 +311,14 @@ export class Pinecone {
    * @throws {@link Errors.PineconeArgumentError} when invalid arguments are provided
    */
   deleteIndex(indexName: IndexName) {
-    return this._deleteIndex(indexName);
+    const deleteIndexPromise = this._deleteIndex(indexName);
+
+    // When an index is deleted, we need to evict the host from the IndexHostSingleton cache.
+    deleteIndexPromise.then(() => {
```
austin-denoble, line 317: I believe this should work for evicting things from the cache if the index has been deleted.

src/data/hostUrlSingleton.ts
austin-denoble: Yup, primarily we're satisfying flow. Maybe we could opt to update the OpenAPI spec to make this a bit easier for us.

src/data/types.ts
```diff
@@ -3,9 +3,8 @@ import type { FetchAPI } from '../pinecone-generated-ts-fetch';
 
 export const PineconeConfigurationSchema = Type.Object(
   {
-    environment: Type.String({ minLength: 1 }),
     apiKey: Type.String({ minLength: 1 }),
-    projectId: Type.Optional(Type.String({ minLength: 1 })),
+    hostUrl: Type.Optional(Type.String({ minLength: 1 })),
```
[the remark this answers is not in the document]
austin-denoble: I went ahead and split things apart. We now have:

- `PineconeConfiguration`
- `IndexConfiguration` (which extends `PineconeConfiguration`)

Let me know if this doesn't make sense.

src/pinecone.ts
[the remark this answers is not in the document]
austin-denoble: I do remember the reason: we're returning promises and I went with passing an explicit callback rather than using `.then()` on the return. Using the promise itself and keeping things isolated to `pinecone.ts`.

src/data/hostUrlSingleton.ts
```diff
@@ -0,0 +1,89 @@
...
+      queryParamsStringify,
+      headers: {
+        'User-Agent': buildUserAgent(false),
+      },
+      fetchApi: getFetch(config),
+      middleware,
+    };
+
+    return new IndexOperationsApi(new ApiConfiguration(apiConfig));
+  };
```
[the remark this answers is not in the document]
austin-denoble: It didn't seem like it necessitated a class as we're not doing the same sort of logic as in `vectorOperationsProvider`. We now have a centralized place for the index API configurations: `indexOperationsBuilder`.

src/data/hostUrlSingleton.ts
austin-denoble: Added. I almost felt like we could leave `PINECONE_ENVIRONMENT` as an optional env variable for now and then either use `PINECONE_CONTROLLER_HOST` if provided, or `PINECONE_ENVIRONMENT` to build it if provided, or default to `https://api.pinecone.io`, but that seemed like it would overly complicate things when we can just provide the URL anyways.

PR #153, opened 2023-11-01
Update pinecone-generated-ts-fetch code along with associated types and unit tests

austin-denoble:
## Problem
We need to update the generated client code (`pinecone-generated-ts-fetch`) to align with upcoming API changes.

## Solution
- Regenerate and replace the contents of `/pinecone-generated-ts-fetch`.
- Update relevant types and unit tests, primarily `createIndex` and `listIndexes`.
- Add new unit tests to the `createIndex.validation` suite focused on new required fields in  `CreateIndexOptions`.
- **Bonus:** Fix for `vectorOperationsProvider` - TS wasn't properly inferring types within `buildVectorOperationsConfig`, and we were passing `undefined` to `indexConfigurationParameters.basePath`.

## Type of Change


src/control/createIndex.ts
```diff
@@ -25,6 +28,21 @@ export type CreateIndexOptions = {
...
 
+  /**
+   * The region where you would like your index to be created
+   */
+  region: string;
+
+  /**
+   * The public cloud where you would like your index hosted
+   */
+  cloud: 'gcp' | 'aws' | 'azure';
```
austin-denoble, line 39: This is aligned with the `CreateRequestCloudEnum` in the generated code: https://github.com/pinecone-io/pinecone-ts-client/pull/153/files#diff-678ee0ec5845c7d940569446c8f0a4210f8e23c95500729b0788439fd9759749R124

Maybe it makes more sense to just export and reuse that directly?

src/control/listIndexes.ts
```diff
@@ -1,28 +1,10 @@
...
-/**
- * A partial description of indexes in your project.
- *
- * For full information about each index, see
- * { @link Pinecone.describeIndex }
- */
-export type PartialIndexDescription = {
-  /** The name of the index */
-  name: string;
-};
```
austin-denoble, line 12: My assumption is now that the intent is for the API to return the full objects, we don't want to keep the partial type around. Let me know if that's not the case.

src/control/listIndexes.ts
```diff
@@ -1,28 +1,10 @@
...
+  return async (): Promise<Array<IndexMeta>> => {
+    const response = await api.listIndexes();
 
-    // We know in a future version of the API that listing
-    // indexes should return more information than just the
-    // index names. Mapping these results into an object
-    // will allow us us to add more information in the future
-    // in a non-breaking way.
-    return names.map((n) => ({ name: n }));
+    return response.databases || [];
```
austin-denoble, line 8: I tried messing with the spec, but this `||` is mainly to handle the case where the response is an empty array. I could not get the OpenAPI spec to handle that and just return the empty array rather than `undefined`, so we basically do that here so we don't have to handle an optional response.

Open to feedback.

src/control/types.ts
```diff
@@ -26,6 +26,13 @@ export const ReplicasSchema = positiveInteger;
 export const PodsSchema = positiveInteger;
 export const MetricSchema = nonemptyString;
 export const DimensionSchema = positiveInteger;
+export const RegionSchema = nonemptyString;
+export const CloudSchema = Type.Union([
+  Type.Literal('gcp'),
+  Type.Literal('aws'),
+  Type.Literal('azure'),
+]);
```
austin-denoble, line 34: There might be a better way of doing this, open to feedback. This also isn't handled gracefully by `errorFormatter` in `validator.ts`. I'd like to cut a ticket and update if using literals here is deemed the way to go.

src/data/vectorOperationsProvider.ts
```diff
@@ -37,9 +37,9 @@ export class VectorOperationsProvider {
     return this.vectorOperations;
   }
 
-  buildVectorOperationsConfig(config) {
+  buildVectorOperationsConfig(config: IndexConfiguration) {
     const indexConfigurationParameters: ConfigurationParameters = {
-      basePath: config.hostUrl,
+      basePath: config.indexHostUrl,
```
austin-denoble, line 42: I had updated the name of this field before merging the previous PR (https://github.com/pinecone-io/pinecone-ts-client/pull/148), and we apparently didn't have any type safety here as `config` seemed to be `any`. The provider was essentially giving `basePath: undefined` to each instance of `Configuration`. 🤦‍♂️

src/control/__tests__/createIndex.validation.test.ts
```diff
@@ -79,13 +108,128 @@ describe('createIndex argument validations', () => {
...
+          name: 'index-name',
+          dimension: 10,
+          region: 'us-east3',
+          cloud: 123,
+          capacityMode: 'pod',
+        });
+
+      expect(toThrow).rejects.toThrowError(PineconeArgumentError);
+      expect(toThrow).rejects.toThrowError(
+        'The argument to createIndex accepts multiple types. Either 1) 2) 3)'
```
austin-denoble, line 183: This is the `errorFormatter` issue mentioned below (`/src/control/types.ts`): https://github.com/pinecone-io/pinecone-ts-client/pull/153/files#diff-cf36f0ee9dc687ce8159fbecd0992953a0c2e646eac44cca76cc2a2ba8246db3R30

We haven't used typebox literals anywhere else and the `errorFormatter` doesn't play nice.

src/control/types.ts
austin-denoble: The specific changes to `validator.ts` to properly handle errors related to this schema are addressed here: https://github.com/pinecone-io/pinecone-ts-client/pull/154

src/control/createIndex.ts
jhamon: It would be nice to adjust the spec to somehow export these directly in the compiled code. But this seems okay for now.

src/control/listIndexes.ts
jhamon: Yes, exactly. This was just a placeholder until we could return the full object.

src/control/listIndexes.ts
jhamon: Even if the spec generation stuff were easier to massage, it seems smart to be defensive like this.

src/data/vectorOperationsProvider.ts
jhamon: Nice to see some type safety added here. But the `index` part of the `indexHostUrl` name seems a bit redundant if the class is called `IndexConfiguration`

jhamon (approved):
This all looks pretty reasonable to me. I haven't looked at your other PR about validation / enums / literals but this seems fine to merge into spruce and continue iterating on in follow-up PRs.

PR #154, opened 2023-11-03
Update validator to handle typebox literal values

austin-denoble:
_This revision is dependent on a previous branches / PRs. These will need to be reviewed and merged first:_ 
- https://github.com/pinecone-io/pinecone-ts-client/pull/153

## Problem
When replacing the OpenAPI generated core in a previous revision (https://github.com/pinecone-io/pinecone-ts-client/pull/153), our `validator` module which handles errors emitted from typebox was not handling literal types correctly.

We added new union of literal types to support `cloud` values:
```
export const CloudSchema = Type.Union([
  Type.Literal('gcp'),
  Type.Literal('aws'),
  Type.Literal('azure'),
]);
```

When passing an incorrect value for the `cloud` field, validator would not handle things gracefully:
Screenshot 2023-11-03 at 7 28 21 PM

The errors emitted from typebox look like this:
Screenshot 2023-11-03 at 7 31 11 PM

## Solution
This took a lot of trial and error and testing, but I think this is a reasonable approach. I'm sure there's a lot more bike shedding and refactoring we could do in `validator`, but for now it feels most sensible to try and cater to the typebox schemas we support.

Previously, at the top level of `errorFormatter` we were pulling out all errors that contained `anyOf` in the schema. We return early with the assumption that these errors are specific to the shape of the top level arguments being validated. Generally, these types of errors will not have an `instancePath`. It is also possible to have `anyOf` errors relating to a specific property, which in our case is the `cloud` field on `createIndex`.

- At the top-level of `errorFormatter()` rename `anyOfErrors` to `anyOfArgumentErrors`, and update the filter to exclude errors with keyword of `const` or `type`.
- Update `typeErrors()` to handle possible `anyOf` errors with keyword `const` and `instancePath.length > 1` as these are errors specific to a property whose type is some set of literal values.
- Update broken unit tests.

## Type of Change

## Test Plan
Pull this PR down and test using the repl:

```
$ npm run repl
$ await init()

// call createIndex and test that passing invalid cloud values throws correctly
$ await client.createIndex({ name: 'test-index', dimension: 10, cloud: 'gg', region: 'us-east4', capacityMode: 'pod' })

// Uncaught:
// PineconeArgumentError: The argument to createIndex had type errors: property 'cloud' is a constant which must be equal to one of: 'gcp', 'aws', 'azure'.
//     at /Users/austin/workspace/pinecone-ts-client/dist/validator.js:239:19
//     at /Users/austin/workspace/pinecone-ts-client/dist/control/createIndex.js:67:21
//     at step (/Users/austin/workspace/pinecone-ts-client/dist/control/createIndex.js:33:23)
//     at Object.next (/Users/austin/workspace/pinecone-ts-client/dist/control/createIndex.js:14:53)
//     at /Users/austin/workspace/pinecone-ts-client/dist/control/createIndex.js:8:71
//     at new Promise (<anonymous>)
//     at __awaiter (/Users/austin/workspace/pinecone-ts-client/dist/control/createIndex.js:4:12)
//     at Pinecone._createIndex (/Users/austin/workspace/pinecone-ts-client/dist/control/createIndex.js:62:40)
//     at Pinecone.createIndex (/Users/austin/workspace/pinecone-ts-client/dist/pinecone.js:243:21)
```

### DIFF
--- a/src/control/listIndexes.ts
+++ b/src/control/listIndexes.ts
@@ -1,5 +1,4 @@
 import { IndexOperationsApi } from '../pinecone-generated-ts-fetch';
-import type { IndexListMeta } from '../pinecone-generated-ts-fetch';
 
 /**
  * A description of indexes in your project.
@@ -27,7 +26,7 @@ export type IndexListDescription = {
 export type IndexList = Array<IndexListDescription>;
 
 export const listIndexes = (api: IndexOperationsApi) => {
-  return async (): Promise<Array<IndexListMeta>> => {
+  return async (): Promise<IndexList> => {
     const response = await api.listIndexes();
 
     return response.databases || [];
--- a/src/validator.ts
+++ b/src/validator.ts
@@ -123,7 +123,7 @@ const typeErrors = (
       const constValueErrors = propErrorGroups[property];
 
       typeErrorsList.push(
-        `property '${property}' is a constant which must be equal to one of: ` +
+        `property '${property}' must be equal to one of: ` +
           Object.values(constValueErrors)
             .map((group) => `'${group.params.allowedValue}'`)
             .join(', ')
--- a/src/validator.ts
+++ b/src/validator.ts
@@ -97,12 +97,46 @@ const typeErrors = (
   messageParts: Array<string>
 ) => {
   const typeErrorsList: Array<string> = [];
+  const anyOfConstPropErrors: Array<ErrorObject> = errors.filter(
+    (error) =>
+      error.schemaPath.indexOf('anyOf') > -1 &&
+      error.keyword === 'const' &&
+      error.instancePath.length > 0
+  );
   let errorCount = 0;
 
+  // handle possible literal types first
+  const propErrorGroups: { [key: string]: Array<ErrorObject> } = {};
+  if (anyOfConstPropErrors.length > 0) {
+    for (const error of anyOfConstPropErrors) {
+      const constValue = error.instancePath.slice(1);
+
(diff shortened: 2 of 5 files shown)


src/validator.ts
```diff
@@ -97,12 +97,46 @@ const typeErrors = (
...
+        propErrorGroups[constValue] = [error];
+      }
+    }
+    const properties = Object.keys(propErrorGroups);
+
+    properties.forEach((property) => {
+      const constValueErrors = propErrorGroups[property];
+
+      typeErrorsList.push(
+        `property '${property}' is a constant which must be equal to one of: ` +
```
jhamon, line 126: I think we could make this slightly more concise.

```suggestion
        `property '${property}' must be equal to one of: ` +
```

jhamon (approved):
Thanks for crawling through the rats nest to try to make this better.

src/control/listIndexes.ts
```diff
@@ -27,7 +26,7 @@ export type IndexListDescription = {
 export type IndexList = Array<IndexListDescription>;
 
 export const listIndexes = (api: IndexOperationsApi) => {
-  return async (): Promise<Array<IndexListMeta>> => {
+  return async (): Promise<IndexList> => {
```
austin-denoble, line 29: Snuck this update in after merging the last PR.

src/validator.ts
austin-denoble: Good idea, thanks!

PR #155, opened 2023-11-06
Add @spruce release to user-agent

austin-denoble:
_This revision is dependent on a previous branches / PRs. These will need to be reviewed and merged first:_ 
- https://github.com/pinecone-io/pinecone-ts-client/pull/153
- https://github.com/pinecone-io/pinecone-ts-client/pull/154

## Problem
We want to be able to track usage specific to the spruce release of the client.

## Solution
- Update `buildUserAgent` logic to check for `packageInfo.release`, and append to the user agent if present.
- Update `release-spruce-dev` and `release-spruce` workflows to add `@spruce` to `version.json` during these specific releases.

## Type of Change

## Test Plan
We will need to release through either the `release-spruce` or `release-spruce-dev` workflows and validate the `version.json` file has the proper info, and the `release` is appended to the user agent header.

### DIFF
--- a/.github/workflows/release-spruce-dev.yml
+++ b/.github/workflows/release-spruce-dev.yml
@@ -34,10 +34,11 @@ jobs:
           currentVersion=$(jq -r '.version' package.json)
           timestamp=$(date +%Y%m%d%H%M%S)
           devVersion="$currentVersion-spruceDev.$timestamp"
+          release="@spruceDev"
           jq --arg newVersion "$devVersion" '.version = $newVersion' package.json > package.tmp && mv package.tmp package.json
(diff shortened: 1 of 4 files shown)


jhamon (approved):
Instead of reading from packageInfo, I would prefer to read an env var that we can set in CI during the build/release process. That way we don't end up with residual code stuff that is spruce-specific when we merge to `main` later. You can ship this as is if you want, but if you do, you should make a ticket to clean it up later.

austin-denoble:
> Instead of reading from packageInfo, I would prefer to read an env var that we can set in CI during the build/release process. That way we don't end up with residual code stuff that is spruce-specific when we merge to `main` later. You can ship this as is if you want, but if you do, you should make a ticket to clean it up later.

That does make more sense. I'll get that changed in this PR.

jhamon (approved):

PR #165, opened 2023-11-15
Guard against empty hostUrl values in IndexHostSingleton cache

austin-denoble:
## Problem
We use `indexHostSingleton` to cache API keys and index `host` addresses to avoid excessive `describeIndex` calls for determining an indexes address. Currently, `IndexHostSingleton._set()` is able to push an empty `''` `host` value into the cache. 

If an index is created and described quickly, this could lead to a situation where `host` is empty, which would break subsequent client calls to that index.

## Solution
Guard against empty `host` values being passed to the cache.

- Throw a `PineconeUnableToResolveHostError` inside `_set` in the event of an empty `hostUrl`
- Add unit test

## Type of Change

## Test Plan
This one is tricky to test as you'd need a scenario where you call `client.CreateIndex({ ... })` followed quickly by `client.describeIndex('index')` before `host` is established. For now, testing that the client works as expected should be sufficient.

### DIFF
--- a/src/data/indexHostSingleton.ts
+++ b/src/data/indexHostSingleton.ts
@@ -66,9 +66,7 @@ export const IndexHostSingleton = (function () {
     ) => {
       // prevent adding an empty hostUrl to the cache
       if (hostUrl === '') {
-        throw new PineconeUnableToResolveHostError(
-          'An invalid host URL was encountered. Please make sure the index exists and is in a ready state.'
-        );
+        return;
       }
 
       const cacheKey = key(config, indexName);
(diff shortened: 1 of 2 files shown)


src/data/indexHostSingleton.ts
```diff
@@ -12,7 +13,7 @@ export const IndexHostSingleton = (function () {
 
   const _describeIndex = async (
     config: PineconeConfiguration,
-    indexName: string
+    indexName: IndexName
   ): Promise<string> => {
     if (!indexOperationsApi) {
       indexOperationsApi = indexOperationsBuilder(config);
```
austin-denoble, line 19: The `if (!host) { ... }` line below on ln 25 should be sufficient from preventing empty host values when the `_describeIndex` function is used internally.

jhamon (approved):
Fix and ship

PR #167, opened 2023-11-21
[spruce] Update control plane operations

austin-denoble:
## Problem
The schema for control plane operations has changed significantly. We need to update all relevant methods, types, and runtime schemas to align with the new client core.

## Solution
For now the focus is on getting a working client, unit tests, TS compilation, and everything other than integration tests passing.

- Updated all relevant control plane operations to support the new request and response shapes.
- Key files to review are under `/src/control/*`. Tests and all of the control modules. The primary differences are in the shape of request and response objects.
- Updated the typebox runtime validation schemas for `configureIndex` and `createIndex` to align with the new request shapes. There's some jank in the errors that are thrown, we will need to make changes to `validator.ts` (see `createIndex.validation.test.ts` or `configureIndex.validation.test.ts`). For now the test files for the validation on these 2 areas should be passing, I've added TODOs.
- Changes to integration tests were pulled in from my previous branch and are mostly still relevant to get our builds passing. For now these are still failing in CI and will be handled in a follow up.
- Added `additionalHeaders` as a field in `PineconeConfiguration`. This is to support interacting with staging environments.

### Follow Up
I've left some TODO comments throughout. I think there will need to be additional passes and refactoring, but in terms of control plane and data plane operations, this should support the new spec.

Key Follow-up items:
- Functional integration tests.
- Audit the generated types we're exporting. Confirm whether we want to "put lipstick on the pig" in certain areas or leave things as-is.
- Update `validator.ts` to handle the new more nested shape of certain arguments.
- Expand validation testing for creating configuring indexes.

## Type of Change

## Test Plan
To validate changes in staging, you'll need to pass a specific header in via `PineconeConfiguration`.

Use the repl to validate control and data plane operations once the correct configuration has been applied:

```
$ npm run repl
$ const client = new Pinecone({ apiKey: <your-api-key>, controllerHostUrl: <your-staging-url>, additionalHeaders: { 'x-environment': <your-dev-environment>}})

$ await client.listIndexes()
$ await client.createIndex({ .... })
...
```


.github/workflows/pr.yml
```diff
@@ -35,123 +35,6 @@ jobs:
       - name: TypeDoc
         uses: ./.github/actions/build-docs
 
-  run-integration-tests:
```
austin-denoble, line 38: This was part of the integration tests work which is also included here. This block of logic was moved to `testing.yml` so it could be more easily reused.

.github/workflows/release-dev.yml
```diff
@@ -6,12 +6,16 @@ on:
     - cron: '0 0 * * *'
 
 jobs:
+  testing:
```
austin-denoble, line 9: I think our Python client runs through testing on the release workflows, felt like it made sense to add it, but let me know otherwise.

.github/workflows/release-prod.yml
```diff
@@ -19,20 +19,18 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # integration-tests:
```
austin-denoble, line 22: Seemed old, we can parameterize the running of tests if desired.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,143 @@
+name: Testing
```
austin-denoble, line 1: This is the logic that was pulled out of `pr.yml`.

src/data/types.ts
```diff
@@ -12,6 +12,8 @@ export const PineconeConfigurationSchema = Type.Object(
     // But declaring it here as Type.Any() is needed to avoid getting caught
     // in the additionalProperties check.
     fetchApi: Type.Optional(Type.Any()),
+
+    httpHeaders: Type.Optional(Type.Any()),
```
austin-denoble, line 16: Allowing custom headers for interacting with staging.

src/index.ts
```diff
@@ -5,21 +5,13 @@ export * as Errors from './errors';
 
 // Type exports
 export type {
-  CollectionDescription,
```
austin-denoble, line 8: Most of the removals here were in lieu of types we're now relying on from the generated core. I think we probably need a more thorough pass of what we're exporting though.

src/utils/testHelper.ts
```diff
@@ -1,11 +0,0 @@
-import { ResponseError } from '../pinecone-generated-ts-fetch';
```
austin-denoble, line 1: I don't think this was being used anywhere.

src/pinecone-generated-ts-fetch/models/InlineResponse200.ts
jhamon, line 1: Good riddance 😂

jhamon (approved):
Thanks for soldiering through. This seems plenty good for the moment.

PR #169, opened 2023-11-30
[spruce] Update control plane integration tests

austin-denoble:
## Problem
Integration tests have been broken for a while now on the spruce branch. We've gotten to the point where we can successfully update control plane tests to validate operations against the new schema.

## Solution
Update the following integration tests to harmonize with the new OpenAPI spec and control plane backend:
- `configureIndex.test.ts`
- `createIndex.test.ts`
- `describeIndex.test.ts`
- `listIndexes.test.ts`
- `errorHandling.test.ts`
- `initialization.test.ts`

For now I've skipped tests under `/integration/data/`. These need updates for the new control plane service along with adding in additional coverage. This will be handled in a follow up PR.

There are also changes to the top-level functions in `Pinecone` for `describeIndex()` and `deleteIndex()`. Through fixing integration tests, I realized that the `promise.then(() => ...)` format I was using for attaching behavior to the unresolved promises was swallowing errors. Restructuring the shape of these functions should resolve those issues.

I also made a small change to `handleApiError()`, adding a new optional `url` parameter. When running some of our integration tests in the edge runtime, I noticed some of our error messages were slightly different. It seems like `ResponseError` could end up with an empty `urlList`, which meant some of our error classes (`PineconeAuthorizationError`, `PineconeNotImplementedError`, etc) produce a slightly different message due to the empty `url` field. From testing, it seems like the `context` object passed to the middleware always has the `url` available, so we fall back to that if `responseError.response.url` is empty.

## Type of Change

## Test Plan
Run integration tests locally, or confirm they are passing in CI.

To run locally, pull this branch down and run for both node and edge:
`TEST_ENV=node npx jest src/integration/ -c jest.config.integration-node.js --runInBand --bail`
`TEST_ENV=edge npx jest src/integration/ -c jest.config.integration-node.js --runInBand --bail`

### DIFF
--- a/src/errors/handling.ts
+++ b/src/errors/handling.ts
@@ -9,7 +9,8 @@ export const handleApiError = async (
   customMessage?: (
     statusCode: number,
     rawMessageText: string
-  ) => Promise<string>
+  ) => Promise<string>,
+  url?: string
 ): Promise<Error> => {
   if (e instanceof Error && e.name === 'ResponseError') {
     const responseError = e as ResponseError;
@@ -21,7 +22,7 @@ export const handleApiError = async (
 
     return mapHttpStatusError({
       status: responseError.response.status,
-      url: responseError.response.url,
+      url: responseError.response.url || url,
       message: message,
     });
   } else if (e instanceof PineconeConnectionError) {
--- a/src/pinecone.ts
+++ b/src/pinecone.ts
@@ -201,18 +201,16 @@ export class Pinecone {
    * @param indexName - The name of the index to describe.
    * @returns A promise that resolves to {@link IndexModel}
    */
-  describeIndex(indexName: IndexName) {
-    const describeIndexPromise = this._describeIndex(indexName);
+  async describeIndex(indexName: IndexName) {
+    const indexModel = await this._describeIndex(indexName);
 
     // For any describeIndex calls we want to update the IndexHostSingleton cache.
     // This prevents unneeded calls to describeIndex for resolving the host for vector operations.
-    describeIndexPromise.then((indexModel) => {
-      if (indexModel.host) {
-        IndexHostSingleton._set(this.config, indexName, indexModel.host);
-      }
-    });
+    if (indexModel.host) {
+      IndexHostSingleton._set(this.config, indexName, indexModel.host);
+    }
 
-    return describeIndexPromise;
+    return Promise.resolve(indexModel);
   }
 
   /**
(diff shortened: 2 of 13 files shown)


src/pinecone.ts
```diff
@@ -201,18 +201,16 @@ export class Pinecone {
    * @param indexName - The name of the index to describe.
    * @returns A promise that resolves to {@link IndexModel}
    */
-  describeIndex(indexName: IndexName) {
-    const describeIndexPromise = this._describeIndex(indexName);
+  async describeIndex(indexName: IndexName) {
```
austin-denoble, line 204: This is the change to support interacting with results from the `describeIndex` call in order to push things into `IndexHostSingleton`.

src/pinecone.ts
```diff
@@ -312,15 +310,13 @@ export class Pinecone {
    * @returns A promise that resolves when the request to delete the index is completed.
    * @throws {@link Errors.PineconeArgumentError} when invalid arguments are provided
    */
-  deleteIndex(indexName: IndexName) {
-    const deleteIndexPromise = this._deleteIndex(indexName);
+  async deleteIndex(indexName: IndexName) {
```
austin-denoble, line 313: Same as above but for evicting the host from `IndexHostSingleton` after a successful delete.

src/errors/handling.ts
```diff
@@ -9,7 +9,8 @@ export const handleApiError = async (
   customMessage?: (
     statusCode: number,
     rawMessageText: string
-  ) => Promise<string>
+  ) => Promise<string>,
+  url?: string
```
austin-denoble, line 13: This is the fallback `url` which is used if the `ResponseError` doesn't have it, for whatever reason.

I spent a lot of time trying to track down why this was happening and it felt like something in the generated runtime. The only test suite that was failing was `errorHandling.test.ts`. When running edge vs. node locally and using the debug environment values, the responses were identical.

src/pinecone.ts
jhamon: Another place you can do something similar is in `listIndexes()` since that response now includes the entire index model with host info and not just the index names.

jhamon (approved):
Nice work on this. 🚢

src/pinecone.ts
austin-denoble: Good call, I've added an Asana task to pursue that as a follow up.  https://app.asana.com/0/1205585657100015/1206081872049522/f

PR #173, opened 2023-12-11
[spruce] Update OpenAPI core to support flexibility in createIndex()

austin-denoble:
## Problem
This PR updates the generated client core based on schema changes made here: https://github.com/pinecone-io/pinecone-protos/pull/97

The current OpenAPI schema is not flexible in terms of which arguments must be passed for the `CreateIndexRequest` operation. `PodSpec` has five required fields which we expect to be present on operations that return a full `IndexModel`.

For `CreateIndexRequest` this is restricting as the generated TypeScript output produces `PodSpec` with all fields as required, which is used in `CreateIndexRequest`. This prevents users from being able to leverage backend defaults, and leads to errors when trying to send requests with all of the fields provided.

Screenshot 2023-12-07 at 12 20 59 PM

## Solution
The OpenAPI spec has been updated here: https://github.com/pinecone-io/pinecone-protos/pull/97

- Regenerate the TypeScript client core and replace `pinecone-generated-ts-fetch`. There's a new `CreateIndexRequestSpecPod` type that represents the shape of configuration for pod indexes. `environment` and `podType` should be the only required fields.
- Update `CreateIndexOptionsSchema` to match the runtime validator to the new expected typing.

There was a bit of flakiness in some integration tests with this PR that I wasn't seeing last week, so I've bumped some of the `sleep()` values in certain tests to give us a bit more breathing room in terms of data freshness. Ideally, there should maybe be a wrapper function in `test-helpers` that allows us to easily retry some number of times if we don't get what we expect. I may look into that when I have some time later.

## Type of Change

## Test Plan
Validate CI passes for unit tests and integration tests.

In order to test locally, pull this branch down and run the repl. Validate that you're able to send a `createIndex()` request with only the required parameters: `environment` and `podType`:

```
$ npm run repl
$ await init()

$ await client.createIndex({ name: 'test-index', dimension: 2, spec: { pod: { environment: 'us-west-1-aws', podType: 'p1.x1', pods: 1 }}})

// You should see an error message denoting 503 - service unavailable which is related to environment. Make sure you're not // seeing any error messages about required fields.
```


src/integration/data/delete.test.ts
```diff
@@ -71,7 +71,7 @@ describe('delete', () => {
 
     // Try deleting the record
     await ns.deleteOne('0');
-    await sleep(1000);
+    await sleep(3000);
```
austin-denoble, line 74: Bumping all these, deletes in particular were having a lot of issues.

src/integration/data/query.test.ts
```diff
@@ -71,7 +71,6 @@ describe('query', () => {
 
     const topK = 5;
     const results = await ns.query({ id: '0', topK });
-    console.log('Query Results: ', results);
```
austin-denoble, line 74: Whoops, left this in here previously.

src/integration/data/upsertAndUpdate.test.ts
```diff
@@ -73,7 +73,7 @@ describe('upsert and update', () => {
       values: newValues,
       metadata: newMetadata,
     });
-    await sleep(3000);
+    await sleep(7000);
```
austin-denoble, line 76: Update was causing issues in edge but not necessarily node. I mentioned this in the PR summary, but I'd like to look into a helper function that lets us retry for asserts in a consistent way.

src/integration/test-helpers.ts
```diff
@@ -118,5 +118,8 @@ export const waitUntilRecordsReady = async (
     indexStats = await index.describeIndexStats();
   }
 
+  // Sleeping one final time before returning for a bit more breathing room for freshness
+  await sleep(sleepIntervalMs);
```
austin-denoble, line 122: band-aid solution as query tests were also giving me a lot of grief

jhamon (approved):
LGTM

Issue #174, opened 2023-12-14
Dev dependencies listed as dependencies


SalmonMode:
It looks like the following packages are listed as dependencies, but can be moved to devDependencies:

- @edge-runtime/types
- @types/node
- typescript

This would help immensely reduce the size of the package after installing, as npm will install these inside the package folder under a nested node_modules folder if the parent package doesn't have a compatible version. Since TypeScript itself isn't really used during runtime of most node apps (and TypeScript 5 is now available), this results in an extra ~80 MB being taken up.

I noticed this because I'm using the client in an AWS lambda layer and it made the lambda layer too big to be uploaded for use. Luckily, these packages should only be needed for compilation, so there shouldn't be any complications from moving them to devDependencies.

austin-denoble:
Good catch, thank you for contributing!

PR #175, opened 2023-12-14
Move non-runtime related deps to devDependencies

SalmonMode:
## Problem

The purpose of this change is to significantly reduce the size of the package when installed. The issue is fully described in #174.

## Solution

The solution is to move the non-runtime related dependencies into the `devDependencies` section of the `package.json` file so that npm doesn't try to install them in a nested `node_modules` folder.

## Type of Change

- [x] None of the above: dependency tweaks

## Test Plan

The code compiled and the unit checks passed. The packages moved are either entirely type declaration files (i.e., `<name>.d.ts`) and/or needed only for compilation (and would be likely be included in the parent package anyway if actually needed). That said, if there's any doubts, we'd know for sure if it'd work after pruning the dev dependencies and trying it out.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,17 +9,16 @@
       "version": "1.1.2",
       "license": "Apache-2.0",
       "dependencies": {
-        "@edge-runtime/types": "^2.2.3",
         "@sinclair/typebox": "^0.29.0",
-        "@types/node": "^18.11.17",
         "ajv": "^8.12.0",
         "cross-fetch": "^3.1.5",
-        "encoding": "^0.1.13",
-        "typescript": "^4.9.4"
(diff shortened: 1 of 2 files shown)


austin-denoble (approved):
This looks good, thank you for taking the initiative to clean our dependencies up a bit! 

CI is failing due to the inability to share repo secrets with forks, I posted a draft PR just to validate the TS compile checks and integration tests: https://github.com/pinecone-io/pinecone-ts-client/pull/176

SalmonMode:
Awesome! Glad I could help!

netroy:
I just discovered that this package was making up for over 10% of our docker image and was about to send this same PR 😅 
Thanks for fixing it @SalmonMode 💟

PR #177, opened 2023-12-28
[spruce] Update configureIndex and createCollection to return proper models, add assertWithRetries integration utility

austin-denoble:
## Problem
There's a few different problems this PR is looking to address:
- The semantic search example app we use as a part of testing CI has been bumped to the latest SDK version in a `spruce` branch: https://github.com/pinecone-io/semantic-search-example/pull/10. We can target this branch to get CI passing again.
- The `configureIndex` and `createCollection` operations are now expected to return `IndexModel` and `CollectionModel` respectively which the code is not capturing.
- Our integration tests are still a bit flakey, especially regarding the `query` tests where it seems like freshness via query sometimes lags causing us to have to retry tests manually.

## Solution
- Update the `/workflows/testing` file to use the `spruce` branch for `semantic-search-example`.
- Update `configureIndex` and `createCollection` operations to return the expected models. Tweak relevant unit tests.
- Add a new `assertWithRetries` function in `test-helpers.ts`. This util takes in an async function and an array of assertions, and will retry a set number of times with a specific delay.

## Type of Change

## Test Plan
Check that CI is green across the board.

To run the integration tests which have been updated to use the new utility locally:
`TEST_ENV=node npx jest src/integration/data/query.test.ts -c jest.config.integration-node.js --runInBand --bail`
`TEST_ENV=edge npx jest src/integration/data/query.test.ts -c jest.config.integration-edge.js --runInBand --bail`

### DIFF
--- a/src/integration/test-helpers.ts
+++ b/src/integration/test-helpers.ts
@@ -123,3 +123,29 @@ export const waitUntilRecordsReady = async (
 
   return indexStats;
 };
+
+type Assertion = (result: any) => void;
+
+export const assertWithRetries = async (
+  asyncFn: () => Promise<any>,
+  assertions: Assertion[],
+  maxRetries: number = 5,
+  delay: number = 3000
+) => {
+  let attempts = 0;
+
+  while (attempts < maxRetries) {
+    try {
+      const result = await asyncFn();
+      assertions.forEach((assertion) => assertion(result));
+      return;
+    } catch (error) {
+      attempts++;
+      if (attempts < maxRetries) {
+        await sleep(delay);
+      } else {
+        throw error;
+      }
+    }
+  }
+};
--- a/src/integration/test-helpers.ts
+++ b/src/integration/test-helpers.ts
@@ -141,7 +141,7 @@ export const assertWithRetries = async (
       return;
     } catch (error) {
       attempts++;
-      if (attempts < maxRetries) {
+      if (attempts <= maxRetries) {
         await sleep(delay);
         // Double the delay for exponential backoff
         delay *= 2;
--- a/src/integration/test-helpers.ts
+++ b/src/integration/test-helpers.ts
@@ -143,6 +143,8 @@ export const assertWithRetries = async (
       attempts++;
       if (attempts < maxRetries) {
         await sleep(delay);
+        // Double the delay for exponential backoff
+        delay *= 2;
       } else {
         throw error;
       }
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -120,6 +120,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: pinecone-io/semantic-search-example
+          ref: spruce
           path: semantic-search-example
       - name: Install and build client
         shell: bash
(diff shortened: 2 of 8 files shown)


src/integration/test-helpers.ts
```diff
@@ -123,3 +123,29 @@ export const waitUntilRecordsReady = async (
 
   return indexStats;
 };
+
+type Assertion = (result: any) => void;
+
+export const assertWithRetries = async (
```
rohanshah18, line 129: This is really great, thanks for adding this function. I have sent you some optional notes internally for this function about the numbers.

rohanshah18:
Internally, I have started a discussion on a few points but overall it looks great to me. Just one question, do we plan on updating all of the tests with assert with retry mechanism at some point in the future to guarantee the integration tests will be successful every time they are executed?

rohanshah18 (approved):

austin-denoble:
> Internally, I have started a discussion on a few points but overall it looks great to me. Just one question, do we plan on updating all of the tests with assert with retry mechanism at some point in the future to guarantee the integration tests will be successful every time they are executed?

We could yeah, I just didn't want to touch all of them within this PR. Primarily we've seen flakiness in the `query` test which is why I implemented them there. However, I've noticed `upsertAndUpdate` also has a tendency to fail where the update hasn't propagated by the time we're asserting, so I'm looking to update those as well.

src/integration/test-helpers.ts
```diff
@@ -123,3 +123,29 @@ export const waitUntilRecordsReady = async (
...
+    } catch (error) {
+      attempts++;
+      if (attempts < maxRetries) {
+        await sleep(delay);
+      } else {
+        throw error;
+      }
+    }
+  }
+};
```
jhamon, line 151: It might be nice to adjust the delay exponential backoff style. Five doublings would turn a 3 second delay into a `3*32 = 96` second delay. Total waiting possible would be `(3*(1 + 2 + 4 + 8 + 16 + 32)) = 189 seconds`. Seems reasonable if that's enough to make the test pass consistently.
 
Also, is there an overall jest timeout you need to worry about running into?

```
export const assertWithRetries = async (
  asyncFn: () => Promise<any>,
  assertions: Assertion[],
  maxRetries: number = 5,
  initialDelay: number = 3000
) => {
  let attempts = 0;
  let delay = initialDelay;

  while (attempts < maxRetries) {
    try {
      const result = await asyncFn();
      assertions.forEach((assertion) => assertion(result));
      return;
    } catch (error) {
      attempts++;
      if (attempts < maxRetries) {
        await sleep(delay);
        delay *= 2; // Double the delay for exponential backoff
      } else {
        throw error;
      }
    }
  }
};
```

jhamon (approved):
I like the idea of it, given the present circumstances. Made a suggestion on adjusting delays.

src/integration/test-helpers.ts
austin-denoble: Good call, agreed with increasing the delay each run, I think that can only help us.

As for the jest timeout, I believe we're currently configured at `100000` ms globally: https://github.com/pinecone-io/pinecone-ts-client/blob/757185042c5c4703b5a4736f4e9836fcf1d65c6f/jest.config.js#L15C19-L15C19

I think this should be more than enough. Given I'm checking `attempts < maxRetries` I think there's only 4 doublings. I think bumping up to `attempts <= maxRetries` gives us one more run, and we should still be under the global timeout:

3 seconds
6 seconds
12 seconds
24 seconds
48 seconds
= 93 seconds

100,000 ms = 100 seconds

PR #180, opened 2024-01-08
[spruce] Normalize host URLs with and without protocol prefix

austin-denoble:
## Problem
We want the client to be flexible enough to allow providing `controllerHostUrl` and `indexHostUrl` with or without the `https://` protocol. This is useful because the `describeIndex` response includes host without the prefix, and we're already prepending it there.

## Solution
- Added a new utility method called `normalizeUrl()`.
- Use `normalizeUrl` in `indexOperationsBuilder`, `vectorOperationsProvider`, and `IndexHostSingleton`. Updating `indexOperationsBuilder` and `vectorOperationsProvider` covers the cases where users are providing custom hosts for control or data plane operations. The `IndexHostSingleton` update replaces the old `hostWithHttps` utility, and we normalize the host before storing it in the cache.
- Added unit tests for `normalizeUrl()`.

## Type of Change

## Test Plan
Verify unit tests and CI pass as expected. The client should continue to work as expected when resolving data plane URLs for index operations, etc.

You can verify manually by providing either a `controllerHostUrl` or an `indexHostUrl` either with or without the prefix:

```
const pinecone = new Pinecone({ apiKey: 'your-api-key', controllerHostUrl: '123-456.svc.io' });

// perform control plane operations, url should be normalized

const index = pinecone.index('test-index', 'index-url-123.svc.pinecone.io');

// perform data plane operations, url should be normalized
```

### DIFF
--- a/src/data/indexHostSingleton.ts
+++ b/src/data/indexHostSingleton.ts
@@ -3,6 +3,7 @@ import type { PineconeConfiguration } from './types';
 import type { IndexName } from '../control';
 import { describeIndex, indexOperationsBuilder } from '../control';
 import { PineconeUnableToResolveHostError } from '../errors';
+import { normalizeUrl } from '../utils';
 
 // We use describeIndex to retrieve the data plane url (host) for a given API key
 // and index. We only ever want to call describeIndex a maximum of once per API key
@@ -36,9 +37,6 @@ export const IndexHostSingleton = (function () {
 
   const key = (config, indexName) => `${config.apiKey}-${indexName}`;
 
-  // describeIndex returns host without the protocol, prepend on return from cache
-  const hostWithProtocol = (host) => `https://${host}`;
-
   return {
     getHostUrl: async function (
       config: PineconeConfiguration,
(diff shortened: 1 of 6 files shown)


src/utils/normalizeUrl.ts
```diff
@@ -0,0 +1,11 @@
+export function normalizeUrl(url?: string): string | undefined {
+  if (!url || url.trim().length === 0) {
```
austin-denoble, line 2: The `.trim()` may be a bit overly defensive, but since we're also normalizing user-provided values, I thought it might be good to check.

src/data/indexHostSingleton.ts
```diff
@@ -36,21 +37,24 @@ export const IndexHostSingleton = (function () {
...
     ) {
       const cacheKey = key(config, indexName);
       if (cacheKey in hostUrls) {
-        return hostWithProtocol(hostUrls[cacheKey]);
+        return hostUrls[cacheKey];
       } else {
         const hostUrl = await _describeIndex(config, indexName);
-        hostUrls[cacheKey] = hostUrl;
-        return hostWithProtocol(hostUrl);
+        this._set(config, indexName, hostUrl);
```
austin-denoble, line 50: This felt a bit messier to have duplication on the return rather than just normalizing things before we store it. Also using `this._set()` rather than accessing the cache directly, not sure which would be preferred tbh.

src/utils/__tests__/normalizeUrl.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    expect(normalizeUrl('http://index-name-abcdef.svc.pinecone.io')).toBe(
+      'http://index-name-abcdef.svc.pinecone.io'
+    );
+  });
+
+  test('keeps https:// protocol if present', () => {
+    expect(normalizeUrl('http://index-name-abcdef.svc.pinecone.io')).toBe(
+      'http://index-name-abcdef.svc.pinecone.io'
+    );
+  });
```
jhamon, line 26:
```suggestion
  test('keeps https:// protocol if present', () => {
    expect(normalizeUrl('https://index-name-abcdef.svc.pinecone.io')).toBe(
      'https://index-name-abcdef.svc.pinecone.io'
    );
  });
```

jhamon (approved):
LGTM except for small test issue

src/utils/__tests__/normalizeUrl.test.ts
austin-denoble: Ah good copy pasta catch, thank you!

PR #183, opened 2024-01-10
[spruce] Regenerate OpenAPI core, update query and fetch to return usage / readUnits

austin-denoble:
## Problem
The backend is now support `readUnits` in responses for the `query` and `fetch` operations. We've updated the `vector_service.proto` file to include the new response shapes, and this needs to be pulled into the clients: https://github.com/pinecone-io/pinecone-protos/pull/105

## Solution
- Regenerate `pinecone-generated-ts-fetch` core and drop it in. This includes @haruska's updates to our tags, thus the renaming of `ManagePodIndexesApi` -> `ManageIndexesApi`.
- Add new `OperationUsage` type to `/src/data/types.ts` (open to changing the name `OperationUsage`).
- Update our custom types for `QueryResponse` and `FetchResponse` to include `usage`.

## Type of Change

## Test Plan
You will need to call either `fetch()` or `query()` and validate the field is returned in the response.

```
await client.index('test-dim-2').fetch(['1', '2'])

// {
//   records: {
//     '1': {
//       id: '1',
//       values: [Array],
//       sparseValues: undefined,
//       metadata: undefined
//     },
//     '2': {
//       id: '2',
//       values: [Array],
//       sparseValues: undefined,
//       metadata: undefined
//     }
//   },
//   namespace: '',
//   usage: { readUnits: 1 }
// }

await client.index('test-dim-2').query({ vector: [0.2, 0.3], topK: 5, includeValues: true })

// {
//   matches: [
//     {
//       id: '1',
//       score: 1.00242746,
//       values: [Array],
//       sparseValues: undefined,
//       metadata: undefined
//     },
//     {
//       id: '2',
//       score: 1.00034547,
//       values: [Array],
//       sparseValues: undefined,
//       metadata: undefined
//     }
//   ],
//   namespace: '',
//   usage: { readUnits: 5 }
// }
```


haruska (approved):
Looks great. Thanks for updating to the new interface!

Issue #141, opened 2023-10-17
[Bug] Breaking change for Bun runtime between v1.0.1 and v1.1.1


erhant:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Consider the following code:

```js
const { Pinecone } = require("@pinecone-database/pinecone");
const apiKey = "your-api-key";
const environment = "your-env";
const indexName = "your-index";

async function main() {
  const pinecone = new Pinecone({
    apiKey,
    environment,
  });
  await pinecone.describeIndex(indexName);
  const index = pinecone.index(indexName);

  // some random query vector
  const arr = Array.from({ length: 1536 }, () => Math.random());

  await index.query({
    vector: arr,
    topK: 50,
    includeValues: true, // it is important that this is `true`!
    includeMetadata: true,
  });
}

main().then(() => process.exit(0));
```

When we run this script with `node index.js` we have no problems. However, when we run it with `bun index.js` it throws an error:

```sh
❯ bun index.js
499 |             var _a;
500 |             return __generator(this, function (_b) {
501 |                 switch (_b.label) {
502 |                     case 0:
503 |                         _a = this.transformer;
504 |                         return [4 /*yield*/, this.raw.json()];
                                                 ^
SyntaxError: Unexpected end of JSON input
      at /Users/erhant/fb/tmp/pinecone-bug/node_modules/@pinecone-database/pinecone/dist/pinecone-generated-ts-fetch/runtime.js:504:45
      at step (/Users/erhant/fb/tmp/pinecone-bug/node_modules/@pinecone-database/pinecone/dist/pinecone-generated-ts-fetch/runtime.js:72:17)
      at /Users/erhant/fb/tmp/pinecone-bug/node_modules/@pinecone-database/pinecone/dist/pinecone-generated-ts-fetch/runtime.js:53:52
      at /Users/erhant/fb/tmp/pinecone-bug/node_modules/@pinecone-database/pinecone/dist/pinecone-generated-ts-fetch/runtime.js:47:8
      at new Promise (:1:20)
      at /Users/erhant/fb/tmp/pinecone-bug/node_modules/@pinecone-database/pinecone/dist/pinecone-generated-ts-fetch/runtime.js:43:11
[...]

PR #184, opened 2024-01-11
[spruce] Add bun support in CI

austin-denoble:
## Problem
Bun has been a rapidly emerging toolkit for developing, testing, running, and bundling Javascript and TypeScript. We currently have an outstanding GH issue related to the bun runtime: https://github.com/pinecone-io/pinecone-ts-client/issues/141. It would be nice to add some coverage in CI for running our integration tests with bun as the runner.

@jhamon kicked this work off a while ago and I wanted to make sure we got this included.

## Solution
- Update `/github/workflows/testing.yml` and add a new `config` section to the `integration-tests` matrix. Tests will now be run with `npm` in node and edge, and `bun` with node for bun versions `1.0.4` -> `1.0.7`. 

## Type of Change

## Test Plan
Make sure all the integration tests are 🟢 in CI.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -44,10 +44,10 @@ jobs:
         uses: ./.github/actions/setup
 
       - name: Setup bun
-        if: matrix.config.runner == 'bun'
+        if: matrix.config.bun_version
         uses: oven-sh/setup-bun@v1
         with:
-          bun-version: $ {{ matrix.config.bun_version }}
+          bun-version: ${{ matrix.config.bun_version }}
 
       - name: Run integration tests
         env:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -24,12 +24,17 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      max-parallel: 1
+      max-parallel: 2
       matrix:
-        jest_env: ['node', 'edge']
-        pinecone_env:
-          - prod
-          # - staging
+        config: [
+            # { runner: 'npm', jest_env: 'node' },
+            # { runner: 'npm', jest_env: 'edge' },
+            # { runner: 'bun', jest_env: 'node' },
+            { runner: 'bun', jest_env: 'node', bun_version: '1.0.4' },
+            { runner: 'bun', jest_env: 'node', bun_version: '1.0.5' },
+            { runner: 'bun', jest_env: 'node', bun_version: '1.0.6' },
+            { runner: 'bun', jest_env: 'node', bun_version: '1.0.7' },
+          ]
 
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


.github/workflows/testing.yml
```diff
@@ -24,12 +24,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      max-parallel: 1
+      max-parallel: 2
```
haruska, line 27: why would one choose serial vs parallel here?

haruska (approved):
:tada: Thanks for adding this!

.github/workflows/testing.yml
austin-denoble: I'm actually not sure why we were opting to run serially before. Each test suite creates a specific index and operates on that currently, so risk of collisions or issues in that sense should be low.

I'm guessing you think we should be doing parallel here?

jhamon: The reason was to manage quota usage. The API key tied to this account can only create a certain number of indexes at one time. We can make it more parallel if we up the quota, which is tied to the sdk-node-testing project in our prod org.

.github/workflows/testing.yml
```diff
@@ -24,12 +24,20 @@ jobs:
...
           - prod
           # - staging
+        config:
+          [
+            { runner: 'npm', jest_env: 'node' },
+            { runner: 'npm', jest_env: 'edge' },
+            { runner: 'bun', jest_env: 'node', bun_version: '1.0.4' },
+            { runner: 'bun', jest_env: 'node', bun_version: '1.0.5' },
+            { runner: 'bun', jest_env: 'node', bun_version: '1.0.6' },
+            { runner: 'bun', jest_env: 'node', bun_version: '1.0.7' },
```
jhamon, line 39: The bun team seems to ship all the time. Since I started this work a couple months ago they shipped like 15 more versions are up to `1.0.22`. It's probably not reasonable to test every single one of those separately, but you could choose some representative numbers throughout the range.

https://github.com/oven-sh/bun/releases

jhamon (approved):

.github/workflows/testing.yml
austin-denoble: Good call, also whoa that's a very rapid release cadence. I'll spread the range out a bit more, thanks.

.github/workflows/testing.yml
```diff
@@ -38,21 +47,49 @@ jobs:
...
         if: matrix.pinecone_env == 'staging'
         env:
           CI: true
-          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY_STAGING }}
+          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
           PINECONE_CONTROLLER_HOST: 'https://api-staging.pinecone.io'
-        run: npm run test:integration:${{ matrix.jest_env }}
+        run: ${{ matrix.config.runner }} run test:integration:${{ matrix.config.jest_env }}
+
+  integration-test-cleanup:
```
austin-denoble, line 71: We never actually use the integration cleanup, and it does seem like there are cases where we end up with orphaned indexes after the integration suite runs. For now I'm explicitly triggering this after the entire matrix has completed, which should be safe since it just lists indexes and deletes any which are present.

.github/workflows/testing.yml
austin-denoble: Added one more step in the matrix, we're now hitting bun `1.0.0`, `1.0.5`, `1.0.10`, `1.0.15`, and `1.0.20`.

src/integration/data/delete.test.ts
```diff
@@ -49,39 +49,39 @@ describe('delete', () => {
...
-    if (records) {
-      expect(records['0'].id).toEqual('0');
-      expect(records['0'].values.length).toEqual(5);
-    } else {
-      fail(
-        'Did not find expected records. Fetch result was ' +
-          JSON.stringify(fetchResult)
-      );
-    }
+    const fetchAssertions = [
```
austin-denoble, line 60: We were seeing a lot of flake in the delete tests across various scenarios. Removed the sleeps and went with `assertWithRetries` which seems to help a lot here. This test probably needs some cleanup eventually.

PR #185, opened 2024-01-13
Add new workflow to trigger building and publishing docs, fix release-prod docs step

austin-denoble:
## Problem
We pushed a patch to fix up our npm release page. In doing this I realize the `release-prod` workflow is incorrectly still trying to push the built client docs to the `gh-pages` branch. It should be moving things to the `sdk-docs` repo like in `merge`.

## Solution

- Add new `build-and-publish-docs` workflow to allow triggering a fresh docs build and deploy off main separate from other workflows.
- Update `release-prod` to use this new workflow if the npm release completes. Also uncommented the test re-use with the boolean flag in that workflow as it surprised me it immediately pushed to npm without any other checks.
- Update other workflows to reuse `build-and-publish-docs`.

## Type of Change

## Test Plan
We'll have to test the release workflow by actually releasing, so give that a careful review. The docs workflow can be triggered and tested on its own.

### DIFF
--- a/.github/workflows/build-and-publish-docs.yml
+++ b/.github/workflows/build-and-publish-docs.yml
@@ -1,0 +1,31 @@
+name: 'Build and publish documentation to sdk-docs'
+
+on:
+  workflow_dispatch: {}
+  workflow_call:
+    secrets:
+      SSH_DEPLOY_KEY:
+        required: true
+
+jobs:
+  build-and-deploy-documentation:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup
+        uses: ./.github/actions/setup
+      - name: Generate TypeDoc documentation
+        uses: ./.github/actions/build-docs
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-publish-docs.yml
```diff
@@ -0,0 +1,31 @@
...
+        env:
+          SSH_DEPLOY_KEY: ${{ secrets.SSH_DEPLOY_KEY }}
+        with:
+          source-directory: docs
+          destination-github-username: pinecone-io
+          destination-repository-name: sdk-docs
+          user-email: clients@pinecone.io
+          target-branch: main
+          target-directory: typescript
+          commit-message: 'TypeScript: automated documentation build - pinecone-ts-client merge SHA: ${{ github.sha }}'
```
jhamon, line 31: I think an extra empty line between commit title and body is conventional.

jhamon (approved):
CI configs look good to me. It's nice reviewing again how the docs publishing works. Will be relevant to me on python very soon. 

Good luck with those tests 💀

PR #189, opened 2024-01-14
Update recordCount -> vectorCount

austin-denoble:
## Problem
Collections still return `vector_count` and we've updated the OpenAPI spec.

## Solution

- Regenerate OpenAPI core
- Fix small issue in `merge` workflow


jhamon (approved):
🚢

PR #190, opened 2024-01-15
Update documentation for v2.0.0 release

austin-denoble:
## Problem
We need to update `README.md` and the `v2-migration.md` files for the upcoming release.

## Solution
Update README code examples and copy. Move the migration guide into the md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,9 +92,7 @@ await pc.createIndex({
 
 #### Create a pod-based index with optional configurations
 
-To create a pod-based index, you define `pod` in the `spec` object which contains the `environment` where the index should be hosted, and the `podType` and `pods` size to use. Many optional configuration fields allow greater control over hardware resources and availability. To learn more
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -29,18 +30,18 @@ npm install @pinecone-database/pinecone
...
+There is one piece of configuration required to use the Pinecone client: an API key. This value can be passed using an environment variable or in code through a configuration object. Find your API key in the console dashboard at [https://app.pinecone.io](https://app.pinecone.io).
 
 #### Using environment variables
 
 The environment variables used to configure the client are the following:
 
 ```bash
 PINECONE_API_KEY="your_api_key"
-PINECONE_ENVIRONMENT="your_environment"
+PINECONE_CONTROLLER_HOST="your_environment"
```
gdj0nes, line 41: Is the for the environment? Seems like something most users won't use so maybe we can bury?

austin-denoble: I had that question too, I think it's fine burying.

jhamon (approved):
🚀 🚀 🚀

PR #197, opened 2024-01-25
Add `sourceTag` to `PineconeConfiguration` and `additionalHeaders` to data plane calls

austin-denoble:
## Problem
We want to provide the ability for consumers to include a `source_tag` to identify the source of requests while using the client. Additionally, there are debugging and development scenarios where being able to pass additional headers with network requests would be convenient.

## Solution
- Add `sourceTag` to `PineconeConfiguration`. Update `buildUserAgent()` to take in `PineconeConfiguration`, which should be available at all the points we call to build the agent. If `sourceTag` is provided it will be normalized, and inserted into the `User-Agent` header for all network requests with this format: `source_tag=<normalizeSourceTag>`. 
  - Normalization rules:
  - Lowercase
  - Limit to charset `[a-z0-9_ ]`
  - Trim left/right space
  - Condense all spaces to one space and replace with `_`
- Add `additionalHeaders` to the `Index` constructor. Custom headers can now be provided when targeting a specific index. `Index` now passes these headers to `VectorOperationsProvider` when setting up the API.
- Update `VectorOperationsProvider` to apply `additionalHeaders` to the API `Configuration` when provided.

## Type of Change
New feature, but specific to developer / support usage.

## Test Plan
New unit test files:
- `indexOperationsBuilder.test.ts`
- `user-agent.test.ts`

New unit test for validating `additionalHeaders` is passed as expected, a bit of refactoring in `index.test.ts` which appears messier than it is.

Running locally with `PINECONE_DEBUG=true` to make sure the custom headers / `sourceTag` are passed on requests when provided.

### Passing `integrationId`
```typescript

const pc = new Pinecone({ apiKey: 'your-api-key', sourceTag: 'test source tag' });
const index = pc.Index('my-index');

// Test control + data plane operations
await pc.listIndexes();
await index.upsert(...);
```

### Passing `additionalHeaders` to `Index`
```typescript

const pc = new Pinecone({ apiKey: 'your-api-key', sourceTag: 'test source tag' });
const index = pc.Index('my-index', undefined, { 'x-custom-header': 'header-value' });

// Make requests
await index.upsert(...);
```

### DIFF
--- a/src/utils/user-agent.ts
+++ b/src/utils/user-agent.ts
@@ -20,9 +20,28 @@ export const buildUserAgent = (config: PineconeConfiguration) => {
     userAgentParts.push(`node ${process.version}`);
   }
 
-  if (config.integrationId) {
-    userAgentParts.push(`integrationId=${config.integrationId}`);
+  if (config.sourceTag) {
+    userAgentParts.push(`source_tag=${normalizeSourceTag(config.sourceTag)}`);
   }
 
   return userAgentParts.join('; ');
+};
+
+const normalizeSourceTag = (sourceTag: string) => {
+  if (!sourceTag) {
+    return;
+  }
+
+  /**
+   * normalize sourceTag
+   * 1. Lowercase
+   * 2. Limit charset to [a-z0-9_ ]
+   * 3. Trim left/right spaces
+   * 4. Condense multiple spaces to one, and replace with underscore
+   */
+  return sourceTag
+    .toLowerCase()
+    .replace(/[^a-z0-9_ ]/g, '')
+    .trim()
+    .replace(/[ ]+/g, '_');
 };
--- a/src/utils/user-agent.ts
+++ b/src/utils/user-agent.ts
@@ -1,7 +1,8 @@
 import { isEdge } from './environment';
+import type { PineconeConfiguration } from '../data/types';
 import * as packageInfo from '../version.json';
 
-export const buildUserAgent = () => {
+export const buildUserAgent = (config: PineconeConfiguration) => {
   // We always want to include the package name and version
   // along with the langauge name to help distinguish these
   // requests from those emitted by other clients
@@ -17,6 +18,10 @@ export const buildUserAgent = () => {
   // If available, capture information about the Node.js version
   if (typeof process !== 'undefined' && process && process.version) {
     userAgentParts.push(`node ${process.version}`);
+  }
+
+  if (config.integrationId) {
+    userAgentParts.push(`integrationId=${config.integrationId}`);
   }
 
   return userAgentParts.join('; ');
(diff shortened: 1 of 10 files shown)


ssmith-pc:
This looks directionally in line with what's needed for [PRD Integration Attributes - SDK Attribution](https://www.notion.so/PRD-Integration-Attribution-7df6c0079b95419ab0bb85f7dee7cd80?pvs=4).

Related PR from Python client: https://github.com/pinecone-io/pinecone-python-client/pull/324

src/utils/user-agent.ts
```diff
@@ -19,5 +20,28 @@ export const buildUserAgent = () => {
...
   }
 
+  if (config.sourceTag) {
+    userAgentParts.push(`source_tag=${normalizeSourceTag(config.sourceTag)}`);
+  }
+
   return userAgentParts.join('; ');
 };
+
+const normalizeSourceTag = (sourceTag: string) => {
```
haruska, line 30: is this expected behavior from the customer setting the source tag?

haruska (approved):
Looks good. Is there a corresponding docs (or README) update explaining the new config?

src/utils/user-agent.ts
austin-denoble: Yeah, it's mimicking what we're doing to the slug on other clients:
- https://github.com/pinecone-io/pinecone-java-client/blob/1480814d20be1b5359e806936ccb50c9de60281b/src/main/java/io/pinecone/configs/PineconeConfig.java#L92
- https://github.com/pinecone-io/pinecone-python-client/blob/58ea88cfcadc5914c137198b9a15c760b87edcc0/pinecone/utils/user_agent.py#L13

I'm not sure if this was officially capture but I'll send you the details I've got.

src/utils/user-agent.ts
```diff
@@ -19,5 +20,28 @@ export const buildUserAgent = () => {
...
+   * 1. Lowercase
+   * 2. Limit charset to [a-z0-9_ ]
+   * 3. Trim left/right spaces
+   * 4. Condense multiple spaces to one, and replace with underscore
+   */
+  return sourceTag
+    .toLowerCase()
+    .replace(/[^a-z0-9_ ]/g, '')
+    .trim()
+    .replace(/[ ]+/g, '_');
```
ssmith-pc, line 46: Minor nit -- I I think the regex is overkill here vs the more readable `.replaceAll(' ', '_')`

ssmith-pc (approved):
LGTM overall

src/utils/user-agent.ts
austin-denoble: I think we would need to update the library we're targeting with TypeScript in order to use `replaceAll()`. Also, would that not replace multiple spaces with multiple underscores? 

```javascript
const testString = "   test_string   lol   hehehe"
testString.replaceAll(' ', '_')
// '___test_string___lol___hehehe'
```

ssmith-pc: My bad, you're right

PR #199, opened 2024-01-31
types(update): wrap metadata param in `Partial`

omikader:
## Problem

This PR makes it so that you don't need to supply all of the required fields in your custom metadata interface when updating a vector using the TS client.

For example, say I have a metadata interface defined as the following:

```
interface MyMetadata {
  title: string;
  source: string;
  isArchived?: boolean;
}
```
I want to update the value for `isArchived`, but I want to leave the values for `title` and `source` as is. I'd write the following code, but I get a type error because no value for `title` or `source` were provided

```
const ns = pinecone.index<MyMetadata>(indexName).namespace(namespace);
await ns.update({ id: "abc123", metadata: { isArchived: true });
```

```
Type '{ isArchived: boolean; }' is missing the following properties from type '{ title: string; source: string; isArchived?: boolean | undefined; }': title, source
```

## Solution

By wrapping the generic in `Partial`, we make it so that all of the fields in the interface are optional which is the intent of this method according to the [documentation](https://docs.pinecone.io/reference/update)

> If set_metadata is included, the values of the fields specified in it will be added or overwrite the previous value.

## Type of Change

## Test Plan

N/A

### DIFF
--- a/src/data/update.ts
+++ b/src/data/update.ts
@@ -45,7 +45,7 @@ export type UpdateOptions<T extends RecordMetadata = RecordMetadata> = {
   /**
    * The metadata you would like to store with this record.
    */
-  metadata?: T;
+  metadata?: Partial<T>;
 };
 
 export class UpdateCommand<T extends RecordMetadata = RecordMetadata> {


austin-denoble:
Hey @omikader, thanks for the great PR description and fix. This looks like it makes sense to me, I'll let CI run and merge when ready, thank you!

austin-denoble (approved):
CI checks out here: https://github.com/pinecone-io/pinecone-ts-client/pull/200

Thanks again for your contribution!

PR #203, opened 2024-02-22
Regenerate OpenAPI core

austin-denoble:
## Problem
There have been a number of changes to the OpenAPI spec including renaming of tags, inclusion of the List endpoint. We need to regenerate and replace `pinecone-generated-ts-fetch`.

## Solution
- Regenerate `/pinecone-generated-ts-fetch` off of the most current OpenAPI spec. Everything under `/pinecone-generated-ts-fetch` is auto-generated and can be ignored.
- Rename `VectorOperationsProvider` -> `DataOperationsProvider` to keep things consistent. We moved away from "vector" terminology previously and since our OpenAPI tags have changed this felt like a good time to update our client naming.

## Type of Change

## Test Plan
There are no changes to client functionality in this PR. All unit tests, builds, and integration tests should pass as expected in CI.


src/data/__tests__/deleteAll.test.ts
```diff
@@ -3,13 +3,13 @@ import { setupDeleteSuccess } from './deleteOne.test';
 
 describe('deleteAll', () => {
   test('calls the openapi delete endpoint, passing deleteAll with target namespace', async () => {
-    const { VoaProvider, VOA } = setupDeleteSuccess(undefined);
+    const { DataProvider, DPA } = setupDeleteSuccess(undefined);
```
austin-denoble, line 6: Renamed these across tests to keep things consistent with the new `DataPlaneApi` class.

src/data/dataOperationsProvider.ts
```diff
@@ -13,11 +13,11 @@ import {
 import { IndexHostSingleton } from './indexHostSingleton';
 import { middleware } from '../utils/middleware';
 
-export class VectorOperationsProvider {
+export class DataOperationsProvider {
```
austin-denoble, line 16: Renamed this class and file to `DataOperationsProvider`.

src/data/__tests__/dataOperationsProvider.test.ts
```diff
@@ -1,7 +1,7 @@
-import { VectorOperationsProvider } from '../vectorOperationsProvider';
+import { DataOperationsProvider } from '../dataOperationsProvider';
```
austin-denoble, line 1: See `dataOperationsProvider` file below, renamed to align with the new API class.

src/integration/data/query.test.ts
```diff
@@ -12,9 +12,10 @@ describe('query', () => {
     index: Index,
     ns: Index,
     namespace: string,
-    recordIds: string[];
+    recordIds: string[],
+    numberOfRecords: number;
 
-  beforeEach(async () => {
+  beforeAll(async () => {
```
austin-denoble, line 18: The primary changes to this file are:

- Only seeding record data in the `beforeAll` and using those records across tests rather than deleting and re-upserting every test, which was time consuming and prone to flake.
- Fixing `query with includeValues: true` as it wasn't properly upserting before deleting (whoops). This passed before but I think we started returning 404s for deletes on non-existing records which broke things.

jhamon (approved):
Nice job propagating the VectorServiceApi => DataPlaneApi rename through all the derivative variables and classes.

PR #207, opened 2024-03-06
Fix listPaginated example in README

austin-denoble:
## Problem
Was reviewing the outgoing docs changes for `list` support and realized the code examples in the README were calling `.list()` rather than `.listPaginated()`.

## Solution
Update the README.

## Type of Change

### DIFF
--- a/README.md
+++ b/README.md
@@ -772,7 +772,7 @@ of record ids, this can be used to help model hierarchical relationships between
 ```typescript
 const pc = new Pinecone();
 const index = pc.index('my-index').namespace('my-namespace');
-const results = await index.list({ prefix: 'doc1' });
+const results = await index.listPaginated({ prefix: 'doc1' });
 console.log(results);
 // {
 //   vectors: [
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -772,7 +772,7 @@ of record ids, this can be used to help model hierarchical relationships between
 ```typescript
 const pc = new Pinecone();
 const index = pc.index('my-index').namespace('my-namespace');
-const results = await index.list({ prefix: 'doc1' });
+const results = await index.listPaginated({ prefix: 'doc1' });
```
jseldess, line 775: I've been told that passing a prefix without some special separator can result in some unexpected results in some cases, so we should use a separator, I think:
```suggestion
const results = await index.listPaginated({ prefix: 'doc1#' });
```

austin-denoble: Good callout, thanks!

haruska (approved):

jhamon (approved):

lnhsingh (approved):

PR #228, opened 2024-06-11
Update CI to test newer versions of Node and test fewer versions of Bun

aulorbe:
## Problem

We currently only test Node version `16.x`, but there are much newer versions with upgraded capabilities. To ensure full coverage, we want to add in `18.x` and `20.x`. This PR does that.

Note: this will increase the time it takes for our CI pipeline to complete, but we have discussed, and the increased latency is worth it to cover more modern versions of Node.

We have also **decreased** the number of `bun` versions tested since they have released a major new version `1.1.x`. We originally had many versions out of an abundance of caution leading up to releasing Pinecone serverless. Now that there have been major releases of `bun` cut more recently, it's my opinion that we only really need to test the following: 
- `1.0.0`, which is basically when they GAed `bun` (and we've always tested this initial major release)
- `1.0.36`, the last minor release before `1.1.0`.
- `1.1.11`, the most recent version of `bun`, which works with Windows and fixes some bugs found in the initial `1.1.0` release.

Our CI pipeline still takes a while, but ultimately I think this is better test coverage, and we can do a quick follow-up PR with some hacks to decrease the overall time at some point in the near future.

## Type of Change

---
  - https://app.asana.com/0/0/1207545518662996

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -102,19 +102,19 @@ jobs:
       fail-fast: false
       matrix:
         tsVersion: [
-           '~4.1.0',
-           '~4.2.0',
-           '~4.3.0',
-           '~4.4.0',
-           '~4.5.0',
-           '~4.6.0',
-           '~4.7.0',
-           '~4.8.0',
-           '~4.9.0',
-           '~5.0.0',
-           '~5.1.0',
-           '~5.2.0',
-           # 'latest', - typebox breaks with latest TypeScript, re-enable later
+            '~4.1.0',
+            '~4.2.0',
+            '~4.3.0',
+            '~4.4.0',
+            '~4.5.0',
+            '~4.6.0',
+            '~4.7.0',
+            '~4.8.0',
+            '~4.9.0',
+            '~5.0.0',
+            '~5.1.0',
+            '~5.2.0',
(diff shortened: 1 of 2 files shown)


.github/workflows/testing.yml
```diff
@@ -46,6 +47,9 @@ jobs:
 
       - name: Setup
         uses: ./.github/actions/setup
+        with:
+          pinecone_api_key: ${{ secrets.PINECONE_API_KEY }}
```
austin-denoble, line 51: How were we calling `Setup` without passing the API key previously? I thought it was required. 🤔

austin-denoble (approved):
Seems fine to me, also probably a good idea to expand coverage a bit more.

We'll have to keep an eye on how long the integration test runs are taking, I'm worried we're going to be pushing into the 20-30 minute runtime with this but I'm not sure.

.github/workflows/testing.yml
aulorbe: Hm, looking at the `Setup` action itself (`setup/action.yml`), I don't think we actually need the API key here, because `Setup` really only installs node version and linting and whatnot; it doesn't interact with Pinecone.

Let me try to remove it here...

aulorbe: removed :)

PR #231, opened 2024-06-13
Add workflow and action files for RC code

aulorbe:
## Problem

We would like to publish RC (release candidate) code to `npm`. In order to do that, we need new github workflow and action files. This PR adds those files, so that we can use the RC branch (`2024-07-rc`) to push to `npm` for users.
"
Since versioning RCs is a bit of an antipattern, these workflows do _not_ contain versioning commands. The "version" is controlled by this command, which just appends `.rc`: `run: npm version prerelease --preid=rc --no-git-tag-version`.

TBD on whether we need Github tags, either. Would love reviewers' input on that (I'm thinking we don't).

## Type of Change

### DIFF
--- a/.github/actions/npm-rc-release/action.yml
+++ b/.github/actions/npm-rc-release/action.yml
@@ -1,4 +1,4 @@
-name: 'npm rc release'
+name: 'Npm RC release'
 description: 'Publish code changes to npmjs.org'
 inputs:
   git_email:
@@ -8,12 +8,8 @@ inputs:
     description: 'Commit author name'
     required: true
   npm_token:
-    description: 'npm token for publishing. Login to npmjs.org to generate this.'
+    description: 'Npm token for publishing. Login to npmjs.org to generate this.'
     required: true
-outputs:
-  release_tag:
-    description: 'The name of the created git tag'
-    value: ${{ steps.release-tag-step.outputs.release-tag }}
 runs:
   using: 'composite'
   steps:
@@ -25,20 +21,15 @@ runs:
     - name: 'Bump version'
       shell: bash
       run: npm version prerelease --preid=rc --no-git-tag-version
-    - name: 'Commit and tag version changes'
+    - name: 'Commit changes'
       shell: bash
       run: |
         git add .
         newVersion=$(jq -r '.version' package.json)
-        git commit -m "[skip ci] Publish release v$newVersion"
-        git tag "v$newVersion"
-    - name: 'Output release tag'
-      shell: bash
-      id: release-tag-step
-      run: echo "release-tag=$(git describe --tags)" >> $GITHUB_OUTPUT
+        git commit -m "[skip ci] Publish RC release v$newVersion"
     - name: 'Push changes'
       shell: bash
-      run: git push --follow-tags
+      run: git push
     - name: 'Publish to npm'
       run: npm publish
       shell: bash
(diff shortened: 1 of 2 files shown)


.github/actions/npm-rc-release/action.yml
```diff
@@ -0,0 +1,46 @@
...
+        newVersion=$(jq -r '.version' package.json)
+        git commit -m "[skip ci] Publish release v$newVersion"
+        git tag "v$newVersion"
+    - name: 'Output release tag'
+      shell: bash
+      id: release-tag-step
+      run: echo "release-tag=$(git describe --tags)" >> $GITHUB_OUTPUT
+    - name: 'Push changes'
+      shell: bash
+      run: git push --follow-tags
```
austin-denoble, line 41: I think we should remove any git tagging we're doing here for release candidates. We can probably emit the tag we used to publish in this step if we want to echo it out in the workflow (`run: echo "${{ steps.npm-rc-release.outputs.release_tag }} was published"`), but it's up to you. 🤔

aulorbe: Okay, will do!

aulorbe: (I think if we remove these tagging parts, the reference to `...outputs.release_tag` won't work anymore, but maybe you don't mean to remove all the refs to `release_tag`? LMK!)

.github/actions/npm-rc-release/action.yml
```diff
@@ -0,0 +1,37 @@
...
+        git config --global user.name "${{ inputs.git_username }}"
+    - name: 'Bump version'
+      shell: bash
+      run: npm version prerelease --preid=rc --no-git-tag-version
+    - name: 'Commit changes'
+      shell: bash
+      run: |
+        git add .
+        newVersion=$(jq -r '.version' package.json)
+        git commit -m "[skip ci] Publish RC release v$newVersion"
```
austin-denoble, line 29: I think this will commit changes to whichever branch is targeted by the action. Possibly OK unless it'd be irritating to deal with when trying to port the inference stuff over to the public repo because we'd have these release commits to deal with.

I'm actually not sure if we should be committing or pushing at all, would the workflow work if we just run `npm version prerelease` and then `npm publish`?

austin-denoble (approved):
Thanks for pulling out the tagging! This looks good to me although I'm still wondering about whether we want to push and commit the RC changes at all.

.github/actions/npm-rc-release/action.yml
aulorbe: Okay, after speaking with @austin-denoble and @ssmith-pc offline, we have come to this conclusion, which we are implementing here: 
- if `main` is the latest release, it does not make sense to merge RC version commits into `main`;
- if `main` is the latest _development_ branch, it _does_ make sense to merge RC version commits there

Since our repo falls under the 1st bullet point, we will _not_ be pushing RC version commits to Github.

.github/actions/npm-rc-release/action.yml
```diff
@@ -0,0 +1,28 @@
...
+  using: 'composite'
+  steps:
+    - name: 'Configure git CLI'
+      shell: bash
+      run: |
+        git config --global user.email "${{ inputs.git_email }}"
+        git config --global user.name "${{ inputs.git_username }}"
+    - name: 'Bump version'
+      shell: bash
+      run: npm version prerelease --preid=rc --no-git-tag-version
```
ssmith-pc, line 23: Can we make the "rc" here configurable through the workflow? 

So that someone creating an RC could pass in something like "rc.2024-07" for example

aulorbe: yes!

PR #235, opened 2024-07-01
Add openapi code generation script and api versioning

jhamon:
## Problem

We need a way to generate code from updated openapi specs

## Solution

- Copy & modify overall approach used in python repository for similar purpose
  - Create new `codegen/` directory to be the home for code generation stuff
  - Use git submodules to get information from `apis` repo
  - Build using the `typescript-fetch` generator in the v7 openapi generator cli. This is the same as we have used in the past, to keep the diff somewhat small.
  - Build data and control planes separately and copy outputs into `src/pinecone-generated-ts-fetch` directory in source. This has been the traditional home for these generated files.
- Since we're no longer mashing specs together ahead of code generation, I need to adjust import paths to find either `pinecone-generated-ts-fetch/control` or `pinecone-generated-ts-fetch/data` as appropriate
- In addition to the default generated output, write a constant called `X_PINECONE_API_VERSION` in each generated module. This will be used when building header configuration with `X-Pinecone-Api-Version` for each request.

## Type of Change

## Test Plan

`npm run generate:openapi`


src/control/indexOperationsBuilder.ts
```diff
@@ -25,6 +26,7 @@ export const indexOperationsBuilder = (
     queryParamsStringify,
     headers: {
       'User-Agent': buildUserAgent(config),
+      'X-Pinecone-Api-Version': X_PINECONE_API_VERSION,
```
jhamon, line 29: Here's where we pass the api version header.

codegen/build-oas.sh
```diff
@@ -0,0 +1,81 @@
...
+
+destination="src/pinecone-generated-ts-fetch"
+build_dir="build"
+
+update_apis_repo() {
+	echo "Updating apis repo"
+	pushd codegen/apis
+		git fetch
+		git pull
+		just build
```
austin-denoble, line 16: Leaving this in seems to cause failures when running locally:

```
npm run generate:openapi

> @pinecone-database/pinecone@2.2.2 generate:openapi
> ./codegen/build-oas.sh 2024-07 && npm run build && npm run format

+ version=2024-07
+ modules=("control" "data")
+ destination=src/pinecone-generated-ts-fetch
+ build_dir=build
+ update_apis_repo
+ echo 'Updating apis repo'
Updating apis repo
+ pushd codegen/apis
~/workspace/pinecone-ts-client/codegen/apis ~/workspace/pinecone-ts-client
+ git fetch
+ git pull
Already up to date.
+ just build
error: No justfile found
```

Is the just command still needed here?

codegen/build-oas.sh
```diff
@@ -0,0 +1,81 @@
...
+		exit 1
+	fi
+}
+
+verify_directory_exists() {
+	local directory=$1
+	if [ ! -d "$directory" ]; then
+		echo "Directory does not exist at $directory"
+		exit 1
+	fi
```
austin-denoble, line 44: Tried checking out your branch and running locally, but it seems like there are issues with the submodule and verifying if the directory exists. Not sure if related to the above, or just the fact that I'm trying to do this in a branch:

```
+ echo 'Verifying spec version 2024-07 exists in apis repo'
Verifying spec version 2024-07 exists in apis repo
+ '[' -z 2024-07 ']'
+ verify_directory_exists codegen/apis/_build/2024-07
+ local directory=codegen/apis/_build/2024-07
+ '[' '!' -d codegen/apis/_build/2024-07 ']'
+ echo 'Directory does not exist at codegen/apis/_build/2024-07'
Directory does not exist at codegen/apis/_build/2024-07
+ exit 1
```

austin-denoble (approved):
Nice work, looks good to me overall. I was having trouble when trying to run the build process locally after checking out your branch, left a few comments around that.

codegen/build-oas.sh
austin-denoble: I'm not sure if this is relevant here, but I had to do this when calling justfile commands in the submodule in a parent repo:

`just -f apis/Justfile build`

Issue #218, opened 2024-05-22
[Bug] Cannot preform hybrid search, query by vector does not take sparse vector value


max-tano:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

![Screenshot 2024-05-22 at 3 10 48 PM](https://github.com/pinecone-io/pinecone-ts-client/assets/146136444/cc3590f0-557d-4e4c-9e39-0b16cd57a7a6)

Query by vector values only takes in the RecordValues type and not the RecordSparseValues type shown below
![Screenshot 2024-05-22 at 3 11 52 PM](https://github.com/pinecone-io/pinecone-ts-client/assets/146136444/d67105fe-7cf7-4ef1-9d57-2ec247f448e6)

Im able to do this query in python but does not seem like functionality is in typescript client yet?

### Expected Behavior

I should be able to send in both dense and sparse vector values to preform hybrid search

### Steps To Reproduce

https://docs.pinecone.io/guides/data/query-sparse-dense-vectors

following that but in typescript, using this version    "@pinecone-database/pinecone": "^2.0.1"

### Environment

```markdown
- **OS**:
- **Language version**:
- **Pinecone client version**:
```

max-tano:
Edit can confirm it doesnt work in the latest version 2.2.1

mcpaddy:
Hello @max-tano,

Thank you for raising this issue. I have tried to reproduce it, and it works successfully using `pinecone@2.2.0`.

Here is my code snippet:

```
import { Pinecone } from '@pinecone-database/pinecone';

const pc = new Pinecone({ apiKey: "API KEY" })
const index = pc.index("sparse")

const vectors = [
      {
              id: 'vec1',
              values: [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8],
              sparseValues: {
                          indices: [1, 5],
                          values: [0.5, 0.5]
                      },
              metadata: {genre: 'drama'},
            },
      {
              id: 'vec2',
              values: [0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9],
              metadata: {'genre': 'action'},
              sparseValues: {
                          indices: [5, 6],
                          values: [0.4, 0.5]
                      }
            }
]

await index.upsert(vectors);
console.log(await index.query({topK: 10,
[...]

PR #236, opened 2024-07-03
Add `sparseVector` to `QueryByVectorValues` type

austin-denoble:
## Problem
`sparseVector?: RecordSparseValues` was left out of the `QueryByVectorValues` type. While the API supports sending `sparseVector` with query requests, if you attempt to compile the client with TypeScript you'll get a build error.

Issue: https://github.com/pinecone-io/pinecone-ts-client/issues/218

## Solution

- Update the `QueryByVectorValues` type to include `sparseVector?: RecordSparseValues`. This should allow users to pass a `sparseVector` along with their `query` operation if needed.
- Update `query` docstring + integration tests.

## Type of Change

## Test Plan
Verify CI integration tests pass. I added `sparseVector` to several of our happy-path `query` tests.

Testing this via the repl wasn't sufficient as you could pass the value directly, and you don't get TS build errors.

To properly test you'll need a TypeScript project that uses the Pinecone SDK, and queries with a `sparseVector` in the request object. `tsc` will fail if using the previous client.

### DIFF
--- a/src/data/__tests__/indexHostSingleton.test.ts
+++ b/src/data/__tests__/indexHostSingleton.test.ts
@@ -173,11 +173,6 @@ describe('IndexHostSingleton', () => {
     expect(mockDescribeIndex).toHaveBeenNthCalledWith(1, 'index-1');
     expect(mockDescribeIndex).toHaveBeenNthCalledWith(2, 'index-2');
 
-    console.log(
-      'mockIndexOperationsBuilder',
-      JSON.stringify(mockIndexOperationsBuilder.mock)
-    );
-
     expect(mockIndexOperationsBuilder).toHaveBeenCalledTimes(2);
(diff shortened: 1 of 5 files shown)


jhamon (approved):

src/data/__tests__/indexHostSingleton.test.ts
```diff
@@ -173,11 +173,6 @@ describe('IndexHostSingleton', () => {
     expect(mockDescribeIndex).toHaveBeenNthCalledWith(1, 'index-1');
     expect(mockDescribeIndex).toHaveBeenNthCalledWith(2, 'index-2');
 
-    console.log(
```
austin-denoble, line 176: whoops, must have left this in previously 😅

mcpaddy (approved):
Very nice.

PR #237, opened 2024-07-15
Add new `CreateIndexSpec` types, regenerate from latest OpenAPI spec

austin-denoble:
## Problem
There was an internal discussion recently involving `shards` and `replicas` fields in `PodSpec` which were not marked as required. This was originally done to address build issues with generated code in TypeScript, but is an inaccurate representation of our spec. 

`PodSpec` was addressed in this PR, and we need to regenerate the TypeScript code and fix: https://github.com/pinecone-io/apis/pull/102

## Solution
- Regenerate `pinecone-ts-client-fetch` from the latest OpenAPI spec using `npm run generate:openapi`.
- Add new `TypeScript` types so we can maintain the flexibility of some of these optional parameters inside `IndexSpec` when creating an index: `CreateIndexSpec`, `CreateIndexServerlessSpec`, and `CreateIndexPodSpec`. Previously, we were relying on the generated types for everything in create, but there are specific nuances to what's required and what's not, and this seems like the most reasonable way of addressing it at the client level.
- Small fix in `codegen/build-oas.sh` - we need to run `git checkout main` in the submodule when updating.
  
You can ignore changes under `/pinecone-generated-ts-fetch/*` which is where all the code generated from OpenAPI lives.

## Type of Change

## Test Plan
Run unit and integration tests to validate the `createIndex` operation is still working as expected. We're exporting new types from the package and other than that, the changes here shouldn't be apparent from a UX perspective.


src/control/createIndex.ts
```diff
@@ -26,9 +28,9 @@ import {
 /**
  * @see [Understanding indexes](https://docs.pinecone.io/docs/indexes)
  */
-export interface CreateIndexOptions extends Omit<CreateIndexRequest, 'metric'> {
-  /** The distance metric to be used for similarity search. You can use 'euclidean', 'cosine', or 'dotproduct'. Defaults to 'cosine'. */
-  metric?: CreateIndexRequestMetricEnum;
+export interface CreateIndexOptions extends Omit<CreateIndexRequest, 'spec'> {
+  /** This option specifies how the index should be deployed. */
+  spec: CreateIndexSpec;
```
austin-denoble, line 33: Two things here:

`metric` used to be marked as required which is why we overwrote it to optional. This is no longer necessary. Instead we're overwriting `spec` with our custom types below specifically for creating indexes.

src/control/createIndex.ts
```diff
@@ -37,6 +39,66 @@ export interface CreateIndexOptions extends Omit<CreateIndexRequest, 'metric'> {
   suppressConflicts?: boolean;
 }
 
+/**
```
austin-denoble, line 42: Docstrings from all of these are mostly pulled from our docs repo, etc.

jhamon (approved):
LGTM

PR #238, opened 2024-07-16
Implement Deletion Protection

austin-denoble:
## Problem
`deletionProtection` is part of the control plane spec, but we have not implemented it in the TypeScript SDK.

## Solution
- Update `configureIndex` schema and `options` type to `ConfigureIndexRequest` to properly support the new object structure for making requests to configure an index. We needed `deletionProtection` at the top level, and the new nesting for `spec: { pod : { ... }}`.
- Update `createIndex` schema to validate optional `deletionProtection`.
- Refactor `configureIndex` unit / validation tests.
- Refactor / enable `configureIndex` integration tests - these had been disabled during the serverless launch era and I never re-wrote them. Better late than never.

## Type of Change

## Test Plan

- Test that creating and index without passing `deletionProtection` defaults to `disabled`.
- Confirm that you can configure both pod and serverless index `deletionProtection` options.
- Confirm that trying to delete an index with `deletionProtection: 'enabled'` throws an error.
- Validate we're seeing the `deletionProtection` field in operations such as `listIndexes` and `describeIndex`.

### DIFF
--- a/src/control/createIndex.ts
+++ b/src/control/createIndex.ts
@@ -103,7 +103,7 @@ const CreateIndexOptionsSchema = Type.Object(
     name: IndexNameSchema,
     dimension: DimensionSchema,
     metric: MetricSchema,
-
+    deletionProtection: Type.Optional(Type.String()),
     spec: Type.Object({
       serverless: Type.Optional(
         Type.Object({
--- a/src/control/__tests__/configureIndex.test.ts
+++ b/src/control/__tests__/configureIndex.test.ts
@@ -32,8 +32,7 @@ describe('configureIndex', () => {
     const IOA = { configureIndex: fakeConfigure } as ManageIndexesApi;
 
     const returned = await configureIndex(IOA)('index-name', {
-      replicas: 4,
-      podType: 'p2.x2',
+      spec: { pod: { replicas: 4, podType: 'p2.x2' } },
     });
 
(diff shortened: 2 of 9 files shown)


haruska (approved):
Looks good but there seems to be some failing tests?

src/control/createIndex.ts
```diff
@@ -104,7 +103,7 @@ const CreateIndexOptionsSchema = Type.Object(
     name: IndexNameSchema,
     dimension: DimensionSchema,
     metric: MetricSchema,
-
+    deletionProtection: Type.Optional(Type.String()),
```
jhamon, line 106: Is there a way to communicate it's not any string but specifically `'enabled'` or `'disabled'` with these types? Also, out of curiosity, did you do any testing where you try passing random strings in this field? I'm curious if the API is properly validating it.

jhamon (approved):

src/control/createIndex.ts
austin-denoble: Ah good point, I think there is with the schema stuff, let me take a look.

I did test passing random strings, and the API does seem to validate. I wrote a specific integration test for it: https://github.com/pinecone-io/pinecone-ts-client/pull/238/files#diff-2827e004959e078a377f7fd8a1662a3c3a93109869e4070fd5ad6c52acbd8257R146-R148

PR #239, opened 2024-07-17
Remove support for Node `<18.0.0`

austin-denoble:
## Problem
When implementing inference @aulorbe ran into issues related to large response payloads causing timeouts in integration tests. She tracked this down to being related to the specific version of Node being used to run the integration tests. Specifically, this was a problem with versions `<=16.0.0`.

We currently support Node `>14.0.0`, however `v20.0.0` is the current LTS version. `v14.0.0` was EOL April 30, 2023, and `v16.0.0` in September 11th, 2023. Details in this Slack thread: https://pinecone-io.slack.com/archives/C05MYEX7T2L/p1718202858574909?thread_ts=1718135375.511479&cid=C05MYEX7T2L

## Solution
Given the issues we've run into testing on older versions of Node, and these versions being fully EOL'd, we should bump our engine support to `v18.0.0` for now. Details in Slack: https://pinecone-io.slack.com/archives/C05MYEX7T2L/p1718202858574909?thread_ts=1718135375.511479&cid=C05MYEX7T2L

- Update `package.json` to support Node `>=18.0.0`.
- Remove `v16.0.0` from integration testing matrix.

## Type of Change

## Test Plan
`npm run build`
`npm run test:integration`

Validate we've got green across the board for inference integration tests in CI.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -18,6 +18,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Setup
         uses: ./.github/actions/setup
       - name: Prettier
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,13 +24,15 @@ jobs:
         run: npm run format
       - name: Check for prettier changes
         run: |
-          git diff --exit-code >/dev/null 2>&1
           git diff
-          EXIT_CODE=$?
-          if [ $EXIT_CODE -ne 0 ]; then
-            echo "Prettier changes detected. Please run 'npm run format' and commit the changes."
-            exit 1
-          fi
+        # run: |
+        #   git diff --exit-code >/dev/null 2>&1
+        #   git diff
+        #   EXIT_CODE=$?
+        #   if [ $EXIT_CODE -ne 0 ]; then
+        #     echo "Prettier changes detected. Please run 'npm run format' and commit the changes."
+        #     exit 1
+        #   fi
         shell: bash
       - name: ESLint
         run: npm run lint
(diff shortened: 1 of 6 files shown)


haruska (approved):

.github/workflows/testing.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         pinecone_env:
           - prod
           # - staging
-        node_version: [16.x, 18.x, 20.x]
+        node_version: [18.x, 20.x]
```
jhamon, line 32: Don't need to do it right now, but we should probably add testing for node 21, 22, etc.

jhamon (approved):

.github/workflows/testing.yml
austin-denoble: Yeah, I think that's a good idea. For right now I'd like to keep the integration runs somewhat fast, and I think covering v18 & v20 at the moment should be sufficient.

.github/workflows/pr.yml
```diff
@@ -23,11 +23,13 @@ jobs:
...
       - name: Check for prettier changes
+        shell: bash
         run: |
-          git diff --exit-code >/dev/null 2>&1
-          EXIT_CODE=$?
+          git diff --exit-code > /dev/null 2>&1 || EXIT_CODE=$?; echo "Exit code: $EXIT_CODE"
           if [ $EXIT_CODE -ne 0 ]; then
             echo "Prettier changes detected. Please run 'npm run format' and commit the changes."
+            echo "Summary of changes: "
+            git diff
```
austin-denoble, line 32: Small tweaks here to make sure the entire snippet runs and logs `git diff` output when there's failure. Was hard for me to debug when it was the `package-lock.json` file causing issues.

.prettierignore
```diff
@@ -1,3 +1,4 @@
 dist
 node_modules
-codegen
\ No newline at end of file
+codegen
+src/pinecone-generated-ts-fetch
```
austin-denoble, line 4: I think this was lost when I was trying to reconcile the different feature branches.

PR #240, opened 2024-07-18
Add console warning when using SDK in browser context

austin-denoble:
## Problem
We recently had an issue logged around problems using the Pinecone SDK from a browser context because we're applying a user-agent to outgoing HTTP requests, and certain browsers don't play nicely with that.

Ultimately, people should not be using the SDK from a browser as it's a large security risk, and exposes their API key if deployed to production. We'd like to add a warning when we detect the SDK is being used from the browser.

## Solution
- Add `isBrowser` function to `utils/environment.ts`.
- Add private helper `_checkForBrowser()` to `Pinecone` which calls `isBrowser` and calls `console.warn` if true. This is run once when constructing the `Pinecone` class.
- Add a unit test to validate the behavior when `window` is present in the global object.
- Use the generated API version default `X_PINECONE_API_VERSION` in the `inferenceOperationsBuilder`.

## Type of Change

## Test Plan
You'll need to run the SDK from within a browser, which would take some setup. The unit test should be good enough to cover the behavior here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -927,7 +927,7 @@ generateQueryEmbeddings().then((embeddingsResponse) => {
 
 ## Productionizing
 
-**Warning:** The Pinecone SDK is intended for server-side use only. Using the SDK within a browser context in a production setting could lead to exposing your API key to third parties. If you have deployed the SDK to production in a browser, please rotate your API keys.
+**Note:** The Pinecone SDK is intended for server-side use only. Using the SDK within a browser context can expose your API key(s). If you have deployed the SDK to production in a browser, please rotate your API keys.
 
 If you are ready to take a JavaScript application to production where raw performance is the overriding concern, you can set the environment variable `PINECONE_DISABLE_RUNTIME_VALIDATIONS="true"` to disable runtime argument validation in the Pinecone client. Runtime validations are used to provide feedback when incorrect method options are provided, for example if you attempt to create an index without specifying a required dimension property.
 
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ await pc.createIndex({
   metric: 'dotproduct',
   spec: {
     pod: {
+      environment: 'us-east4-gcp',
       pods: 2,
       podType: 'p1.x2',
       metadataConfig: {
--- a/README.md
+++ b/README.md
@@ -927,6 +927,8 @@ generateQueryEmbeddings().then((embeddingsResponse) => {
 
 ## Productionizing
 
+**Warning:** The Pinecone SDK is intended for server-side use only. Using the SDK within a browser context in a production setting could lead to exposing your API key to third parties. If you have deployed the SDK to production in a browser, please rotate your API keys.
+
 If you are ready to take a JavaScript application to production where raw performance is the overriding concern, you can set the environment variable `PINECONE_DISABLE_RUNTIME_VALIDATIONS="true"` to disable runtime argument validation in the Pinecone client. Runtime validations are used to provide feedback when incorrect method options are provided, for example if you attempt to create an index without specifying a required dimension property.
 
 These runtime validations are most helpful for users who are not developing with Typescript or who are experimenting in a REPL or notebook-type setting. But once you've tested an application and have gained confidence things are working as expected, you can disable these checks to gain a small improvement in performance. This will have the most impact if your workload is upserting very large amounts of data.
--- a/src/__tests__/pinecone.test.ts
+++ b/src/__tests__/pinecone.test.ts
@@ -2,6 +2,7 @@ import { Pinecone } from '../pinecone';
 import { IndexHostSingleton } from '../data/indexHostSingleton';
 import type { PineconeConfiguration } from '../data';
 import * as utils from '../utils';
+import { warn } from 'console';
 
 const fakeFetch = jest.fn();
 const fakeHost = '123-456.pinecone.io';
@@ -64,6 +65,12 @@ jest.mock('../control', () => {
 });
 
 describe('Pinecone', () => {
+  let originalWindow: typeof globalThis.window;
+
+  beforeEach(() => {
+    originalWindow = global.window;
+  });
+
   describe('constructor', () => {
     describe('required properties', () => {
       test('should throw an error if apiKey is not provided', () => {
@@ -136,11 +143,32 @@ describe('Pinecone', () => {
       });
 
       test('should throw an error if required environment variable is not set', () => {
-        delete process.env.PINECONE_API_KEY;
         expect(() => new Pinecone()).toThrow(
           "Since you called 'new Pinecone()' with no configuration object, we attempted to find client configuration in environment variables but the required environment variables were not set. Missing variables: PINECONE_API_KEY. You can find the configuration values for your project in the Pinecone developer console at https://app.pinecone.io"
         );
       });
+    });
+
+    test('should log a warning if the SDK is used in a browser context', () => {
+      // Mock the window object since our unit tests are not running in a browser context
+      (global as any).window = Object.create(window);
+      Object.defineProperty(window, 'document', {
+        value: {},
+        writable: true,
+      });
+
+      const warnSpy = jest
+        .spyOn(console, 'warn')
+        .mockImplementation(() => null);
+
+      new Pinecone({ apiKey: 'test-api-key' });
+
+      expect(warnSpy).toHaveBeenCalledWith(
+        'Warning: The Pinecone SDK is intended for server-side use only. Using the SDK within a browser context in a production setting could lead to exposing your API key to third parties. If you have deployed the SDK to production in a browser, please rotate your API keys.'
+      );
+
+      // Rstore the original window
+      delete (global as any).window;
     });
   });
 
--- a/src/__tests__/pinecone.test.ts
+++ b/src/__tests__/pinecone.test.ts
@@ -154,7 +154,7 @@ describe('Pinecone', () => {
       new Pinecone({ apiKey: 'test-api-key' });
 
       expect(warnSpy).toHaveBeenCalledWith(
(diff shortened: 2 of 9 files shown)


src/__tests__/pinecone.test.ts
```diff
@@ -136,12 +136,31 @@ describe('Pinecone', () => {
       });
 
       test('should throw an error if required environment variable is not set', () => {
-        delete process.env.PINECONE_API_KEY;
```
austin-denoble, line 139: There's setup and teardown that makes this redundant.

src/models/embeddingsList.ts
```diff
@@ -25,9 +25,9 @@ export class EmbeddingsList
   extends Array<Embedding>
   implements OpenAPIEmbeddingsList
 {
-  model: string;
-  data: Array<Embedding>;
-  usage: EmbeddingsListUsage;
+  model?: string;
+  data?: Array<Embedding>;
+  usage?: EmbeddingsListUsage;
```
austin-denoble, line 30: I think these should be optional to align with the actual interface that's being implemented. Thoughts @aulorbe? https://github.com/pinecone-io/pinecone-ts-client/blob/fd96dcba09aa8751b2d59eb8be9be73380c4adac/src/pinecone-generated-ts-fetch/control/models/EmbeddingsList.ts#L34

haruska (approved):

README.md
```diff
@@ -927,6 +927,8 @@ generateQueryEmbeddings().then((embeddingsResponse) => {
 
 ## Productionizing
 
+**Warning:** The Pinecone SDK is intended for server-side use only. Using the SDK within a browser context in a production setting could lead to exposing your API key to third parties. If you have deployed the SDK to production in a browser, please rotate your API keys.
```
aulorbe, line 930: Total and complete nit, but I'd phrase this as a `Note` rather than a warning. I feel like `warning` implies there's something bad about this behavior, when in fact it's preferred behavior. Might just be me though!

aulorbe: Just to make this a bit more direct, instead of 

> Using the SDK within a browser context in a production setting could lead to exposing your API key to third parties.

I'd say

> Using the SDK within a browser context can expose your API key(s).

src/__tests__/pinecone.test.ts
```diff
@@ -136,12 +136,31 @@ describe('Pinecone', () => {
...
         );
       });
     });
+
+    test('should log a warning if the SDK is used in a browser context', () => {
+      // Mock window simulate browser context
+      const mockWindow = {} as any;
+      global.window = mockWindow;
+
+      const warnSpy = jest
```
aulorbe, line 150: Should this be an error or bug instead of a warn?

src/models/embeddingsList.ts
```diff
@@ -25,14 +25,14 @@ export class EmbeddingsList
...
-  usage: EmbeddingsListUsage;
+  usage?: EmbeddingsListUsage;
 
   constructor(
-    model: string,
-    data: Array<Embedding>,
-    usage: EmbeddingsListUsage
+    model?: string,
+    data: Array<Embedding> = [],
+    usage?: EmbeddingsListUsage
```
austin-denoble, line 35: This change is related to this previous comment: https://github.com/pinecone-io/pinecone-ts-client/pull/240#discussion_r1683159720

Basically, I think what was happening is you had your custom class expecting non-optionals for all of these, but that doesn't align with the spec. There may have been cases where this was causing problems because of how this class is extending and implementing `Array` and the generated `EmbeddingsList`.

src/inference/inference.ts
```diff
@@ -34,36 +34,17 @@ export class Inference {
...
+    const typedParams: EmbedRequestParameters = this._formatParams(params);
+    const typedRequest: EmbedOperationRequest = {
+      embedRequest: {
+        model: model,
+        inputs: typedAndFormattedInputs,
+        parameters: typedParams,
+      },
+    };
+    const response = await this._inferenceApi.embed(typedRequest);
+    return new EmbeddingsList(response.model, response.data, response.usage);
```
aulorbe, line 48: 👍

src/inference/inferenceOperationsBuilder.ts
```diff
@@ -25,7 +26,7 @@ export const inferenceOperationsBuilder = (
     queryParamsStringify,
     headers: {
       'User-Agent': buildUserAgent(config),
-      'X-Pinecone-API-Version': '2024-07',
+      'X-Pinecone-Api-Version': X_PINECONE_API_VERSION,
```
aulorbe, line 29: Nice thanks

src/integration/inference/embed.test.ts
```diff
@@ -15,10 +15,7 @@ describe('Integration Test: Pinecone Inference API embeddings endpoint', () => {
     };
     model = 'multilingual-e5-large';
     const apiKey = process.env.PINECONE_API_KEY || '';
-    pinecone = new Pinecone({
-      apiKey,
-      additionalHeaders: { 'X-Pinecone-API-Version': '2024-07' },
```
aulorbe, line 20: 👍

src/models/embeddingsList.ts
```diff
@@ -25,14 +25,14 @@ export class EmbeddingsList
...
+  model?: string;
   data: Array<Embedding>;
-  usage: EmbeddingsListUsage;
+  usage?: EmbeddingsListUsage;
 
   constructor(
-    model: string,
-    data: Array<Embedding>,
-    usage: EmbeddingsListUsage
+    model?: string,
```
aulorbe, line 33: Should these be optional? I think you said they're mandatory in the actual API, but do we want ppl to be able to use the endpoint without these? Maybe we use a default model on the backend now (I don't think we did before, but maybe!)

src/inference/inference.ts
austin-denoble, line 48: I don't think we should be throwing or logging errors here. If there's an error from making the request, it should bubble up like our other operations.

See this comment for a bit more context on why I changed this: https://github.com/pinecone-io/pinecone-ts-client/pull/240/files#r1683236813. @aulorbe I wasn't able to directly reproduce what the previous error code was trying to handle, but let me know if there's some way of doing that.

src/pinecone.ts
```diff
@@ -115,9 +113,10 @@ export class Pinecone {
     }
 
     this._validateConfig(options);
-
     this.config = options;
 
+    this._checkForBrowser();
```
aulorbe, line 118: nice

src/models/embeddingsList.ts
aulorbe: sounds great !

src/__tests__/pinecone.test.ts
austin-denoble: [Discussed offline] - we want to warn rather than making the SDK unusable. If people want to use the client in a browser locally while experimenting with Pinecone that's fine, but we need to make it clear productionizing is a security risk.

src/models/embeddingsList.ts
austin-denoble: These fields aren't enforcing how people call the API, they're just enforcing what `EmbeddingsList` the class is expecting the response to look like, so it should actually be somewhat aligned with what the API spec defines.

When making `embed` requests things are enforced through this type: https://github.com/pinecone-io/pinecone-ts-client/blob/release-candidate/2024-07/src/inference/inference.ts#L32-L36

src/__tests__/pinecone.test.ts
```diff
@@ -136,12 +136,31 @@ describe('Pinecone', () => {
...
+        .spyOn(console, 'warn')
+        .mockImplementation(() => null);
+
+      new Pinecone({ apiKey: 'test-api-key' });
+
+      expect(warnSpy).toHaveBeenCalledWith(
+        'Warning: The Pinecone SDK is intended for server-side use only. Using the SDK within a browser context in a production setting could lead to exposing your API key to third parties. If you have deployed the SDK to production in a browser, please rotate your API keys.'
```
aulorbe, line 157: I'd change the warning language to match the updated `note` in the README: https://github.com/pinecone-io/pinecone-ts-client/pull/240/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R931

src/models/__tests__/embeddingsList.test.ts
```diff
@@ -22,7 +22,7 @@ describe('EmbeddingsList', () => {
     expect(embeddingsList.model).toEqual(mockModel);
     expect(embeddingsList.data).toEqual(mockEmbeddings);
     expect(embeddingsList.usage).toEqual(mockUsage);
-    expect(embeddingsList.data.values).toEqual(mockEmbeddings.values);
+    expect(embeddingsList.data?.values).toEqual(mockEmbeddings.values);
```
aulorbe, line 25: This soon won't be optional, though, right? A la this Slack convo? https://pinecone-io.slack.com/archives/C06HPKVKUFM/p1721324916903089

Might just want to make it with a `TODO` to come back to later, or at least make a high-priority Asana ticket

src/pinecone.ts
```diff
@@ -558,6 +557,15 @@ export class Pinecone {
     )(options);
   }
 
+  /** @internal */
+  _checkForBrowser() {
+    if (isBrowser()) {
+      console.warn(
+        'Warning: The Pinecone SDK is intended for server-side use only. Using the SDK within a browser context in a production setting could lead to exposing your API key to third parties. If you have deployed the SDK to production in a browser, please rotate your API keys.'
```
aulorbe, line 564: I'd have this language match the updated `note` in the README.

aulorbe (approved):

src/models/__tests__/embeddingsList.test.ts
austin-denoble: (discussed offline): We're not changing the spec for now.

Issue #212, opened 2024-04-17
[Bug] issue with deps `@sinclair+typebox`


nicolasburtey:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

can you update `@sinclair+typebox`

current version has issue I'm running into

```
../../node_modules/.pnpm/@sinclair+typebox@0.29.6/node_modules/@sinclair/typebox/typebox.d.ts:179:306 - error TS2589: Type instantiation is excessively deep and possibly infinite.

179 export type TIndex<T extends TSchema, K extends Key[]> = T extends TRecursive<infer S> ? TIndex<S, K> : T extends TIntersect ? UnionType<Flat<TIndexRestMany<T, K>>> : T extends TUnion ? UnionType<Flat<TIndexRestMany<T, K>>> : T extends TObject ? UnionType<Flat<TIndexRestMany<T, K>>> : T extends TTuple ? UnionType<Flat<TIndexRestMany<T, K>>> : TNever;
```                                                                                                                       

documented here https://github.com/sinclairzx81/typebox/issues/737                                                                                                                                                                   

### Expected Behavior

pass typescript check

### Steps To Reproduce

-

### Environment

```markdown
- **OS**:
- **Language version**:
- **Pinecone client version**:
```

guinaut:
Not much to add.. also seeing this issue.

hstanford:
I've resolved this by overriding the dependency to the latest (`0.32.30` at the time of writing). For pnpm, top level in package.json:
```
  "pnpm": {
    "overrides": {
      "@pinecone-database/pinecone>@sinclair/typebox": "0.32.30"
    }
  }
```

Typescript version 5.4.5

anawishnoff:
Hi all, closing out this issue as it seems to have been resolved. Please feel free to reopen if needed.

Issue #164, opened 2023-11-14
[Bug] Pinecone client doesn't initialize on Cloudfare workers or Vercel 


Blaizzy:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When I try to initialise the client it throws errors

```js
const pinecone = new Pinecone({
                apiKey: env.PINECONE_API_KEY,
                environment: env.PINECONE_ENVIRONMENT,
});
```

Cloudfare Error: 
```
Error compiling schema, function code: const schema2 = scope.schema[2];const schema1 = scope.schema[1];return function validate1(data, {instancePath="", parentData, parentDataProperty, rootData=data}={}){let vErrors = null;let errors = 0;if(!(((typeof data == "number") && (!(data % 1) && !isNaN(data))) && (isFinite(data)))){const err0 = {instancePath,schemaPath:"#/definitions/nonNegativeInteger/type",keyword:"type",params:{type: "integer"},message:"must be integer"};if(vErrors === null){vErrors = [err0];}else {vErrors.push(err0);}errors++;}if((typeof data == "number") && (isFinite(data))){if(data < 0 || isNaN(data)){const err1 = {instancePath,schemaPath:"#/definitions/nonNegativeInteger/minimum",keyword:"minimum",params:{comparison: ">=", limit: 0},message:"must be >= 0"};if(vErrors === null){vErrors = [err1];}else {vErrors.push(err1);}errors++;}}validate1.errors = vErrors;return errors === 0;}
EvalError: Code generation from strings disallowed for this context
    at new Function (<anonymous>)
    at Ajv.compileSchema (file:///private/var/folders/c8/yj5plb7n12j4_dx2t602mjgw0000gn/T/tmp-36161-E863LF5hXPTq/index.js:10324:30)
    at Ajv.inlineOrCompile (file:///private/var/folders/c8/yj5plb7n12j4_dx2t602mjgw0000gn/T/tmp-36161-E863LF5hXPTq/index.js:10380:49)
    at Ajv.resolveRef (file:///private/var/folders/c8/yj5plb7n12j4_dx2t602mjgw0000gn/T/tmp-36161-E863LF5hXPTq/index.js:10374:47)
    at Object.code (file:///private/var/folders/c8/yj5plb7n12j4_dx2t602mjgw0000gn/T/tmp-36161-E863LF5hXPTq/index.js:12188:47)
[...]

PR #249, opened 2024-08-13
Remove Typebox and AVJ to work better with Edge runtimes

aulorbe:
## Problem

There have been a litany of issues related to the incompatibility of our current Typescript client and Edge runtimes (e.g. issue [164](https://github.com/pinecone-io/pinecone-ts-client/issues/164), issue [212](https://github.com/pinecone-io/pinecone-ts-client/issues/212), [this](https://community.pinecone.io/t/error-using-node-js-sdk-on-edge/6337/4) community forum post).

The long and the short of it is that we use two libraries: sinclair/typebox and ajv that to do not work in Edge runtimes. The result is that users need to downgrade their client version in order to use our client. This has the major disadvantage of meaning that these users don't get access to our client's newest features (e.g. inference).

Asana ticket: https://app.asana.com/0/1203260648987893/1208002049018710/f

## Solution

Remove typebox and ajv!

## Approach
- Previously we had custom typebox schema objects declared in `types.ts` files under `control` and `data`. This PR removes those  typebox schemas. We used those typebox schemas to do things like ensure the correct properties were sent with requests like `Create Index`, which needs a particular `Spec` obj with particular fields, etc. This PR maintains that behavior, but through custom validation functions embedded within the client's native objects (e.g. `QueryCommand`, `createIndex`, etc.) themselves.

- Importantly, these internal changes **do not result** in any downstream changes the user should be made aware of -- all exported Pinecone `types` remain untouched in files like `index.ts` under `data` and `control`. 

- This PR also introduces new unit and integration tests to ensure the above-mentioned functionality. It also removes unit and integration tests that simply test whether a particular field's value is of a particular data type. 
    - I chose to remove these tests since they're redundant in Typescript. The thinking here is that: if we're going to write a Typescript client and glean the benefits of using Typescript (rather than Javascript), we need to rely on Typescript's typing power. Otherwise, we should just write a Javascript client. I know this is unideal for non-Typescript users of our client, but the backend API will throw errors at runtime if non-TS users pass in the incorrect datatypes, and this is enough for now. 

- Finally, this PR introduces checks for unknown field properties/names. If a user passes in `diemnsion` instead of `dimension` when creating an index, an error will be thrown that tells the user 1) what field was identified as unknown and 2) a list of all the acceptable fields for that object.

## Type of Change

## Test Plan

CI passes.


.github/workflows/testing.yml
```diff
@@ -111,7 +112,7 @@ jobs:
             '~5.0.0',
             '~5.1.0',
             '~5.2.0',
-            # 'latest', - typebox breaks with latest TypeScript, re-enable later
```
aulorbe, line 114: 🎉 !

src/control/configureIndex.ts
```diff
@@ -4,54 +4,50 @@ import {
...
-
-  return async (
+  const validator = async (
     indexName: IndexName,
     options: ConfigureIndexRequest
-  ): Promise<IndexModel> => {
-    indexNameValidator(indexName);
-    patchRequestValidator(options);
+  ) => {
+    if (options) {
```
aulorbe, line 15: Example of new utility function `ValidateProperties` in new file `utils/validateProperties.ts`, which takes in a list of acceptable property names and identifies unacceptable property names, if passed.

src/control/configureIndex.ts
```diff
@@ -4,54 +4,50 @@ import {
...
-    'The first argument to configureIndex',
-    IndexNameSchema
-  );
-  const patchRequestValidator = buildValidator(
-    'The second argument to configureIndex',
-    ConfigureIndexOptionsSchema
-  );
-
-  return async (
+  const validator = async (
```
aulorbe, line 11: Example of embedded `validator` function within `configureIndex` call that replaces the old `buildValidator` functionality that required `typebox` to check values against schema objs we created in `types.ts`

src/control/createIndex.ts
```diff
@@ -98,52 +85,134 @@ export interface CreateIndexPodSpec {
...
+        }. Valid values are: ${ValidPodTypes.join(', ')}.`
+      );
+    }
+  };
 
   return async (options: CreateIndexOptions): Promise<IndexModel | void> => {
     // If metric is not specified, default to cosine
     if (options && !options.metric) {
-      options.metric = 'cosine';
+      options.metric = IndexModelMetricEnum.Cosine;
```
aulorbe, line 213: Small nit improvement to use the `Enum` instead of hard-coding the `'cosine'` val here.

src/errors/config.ts
```diff
@@ -1,6 +1,6 @@
 import { BasePineconeError } from './base';
 
-const CONFIG_HELP = `You can find the configuration values for your project in the Pinecone developer console at https://app.pinecone.io`;
+const CONFIG_HELP = `You can find the configuration values for your project in the Pinecone developer console at https://app.pinecone.io.`;
```
aulorbe, line 3: nit.

package.json
```diff
@@ -9,7 +9,8 @@
   "contributors": [
     "Jen Hamon (https://github.com/jhamon)",
     "Roie Schwaber-Cohen (https://github.com/rschwabco)",
-    "Austin DeNoble (https://github.com/austin-denoble)"
+    "Austin DeNoble (https://github.com/austin-denoble)",
+    "Audrey Lorberfeld (https://github.com/aulorber)"
```
aulorbe, line 13: 😂 Had to

package.json
```diff
@@ -9,7 +9,8 @@
   "contributors": [
     "Jen Hamon (https://github.com/jhamon)",
     "Roie Schwaber-Cohen (https://github.com/rschwabco)",
-    "Austin DeNoble (https://github.com/austin-denoble)"
+    "Austin DeNoble (https://github.com/austin-denoble)",
+    "Audrey Lorberfeld (https://github.com/aulorbe)"
```
aulorbe, line 13: 😇

src/control/createIndex.ts
```diff
@@ -98,52 +85,134 @@ export interface CreateIndexPodSpec {
...
-);
-
 export const createIndex = (api: ManageIndexesApi) => {
-  const validator = buildConfigValidator(
-    CreateIndexOptionsSchema,
-    'createIndex'
-  );
+  const validator = async (options: CreateIndexOptions) => {
+    if (options) {
+      ValidateProperties(options, [
```
austin-denoble, line 91: While I appreciate the thoroughness in evaluating the keys that were possibly passed, I think maintaining these inline representations of the types keys is going to be prone to problems when the types shift around. Can we more tightly couple the lists that are being passed to validate things with the type? I'd look at using `keysOf` and generics if possible: https://www.executeprogram.com/courses/advanced-typescript/lessons/keyof-generic-constraints

src/control/createIndex.ts
```diff
@@ -98,52 +85,134 @@ export interface CreateIndexPodSpec {
...
+        )}.`
+      );
+    }
+    if (
+      options.metric &&
+      !Object.values(IndexModelMetricEnum).includes(options.metric)
+    ) {
+      {
+        throw new PineconeArgumentError(
+          `Invalid metric value: ${options.metric}. Valid values are: 'cosine', 'euclidean', or 'dotproduct.'`
```
austin-denoble, line 179: We may have some inconsistencies in string formatting in some of these errors. Here we're denoting fields with `''` but we've used backticks elsewhere. Do we want to standardize that?

src/control/deleteCollection.ts
```diff
@@ -1,22 +1,20 @@
...
-    'deleteCollection'
-  );
-
   return async (collectionName: DeleteCollectionOptions): Promise<void> => {
-    validator(collectionName);
-
+    if (!collectionName || collectionName.length === 0) {
+      throw new PineconeArgumentError(
+        'You must pass a non-empty string for `collectionName` in order to delete a' +
+          ' collection'
```
austin-denoble, line 15: nit: Personal preference but I really hate these multi-line strings and would prefer to avoid them if possible. Is this a prettier or linting thing?

src/control/types.ts
```diff
@@ -82,3 +37,18 @@ export type PodType =
   | 'p2.x4'
   | 'p2.x8'
   | string;
+
+export const ValidPodTypes: PodType[] = [
```
austin-denoble, line 41: It's a bit odd that the `PodType` type includes all these values but then also just `string` 🤔

src/data/describeIndexStats.ts
```diff
@@ -60,16 +53,23 @@ export type DescribeIndexStatsOptions = {
...
-    'describeIndexStats'
-  );
+  const validator = async (options: DescribeIndexStatsOptions) => {
+    if (options) {
+      ValidateProperties(options, ['filter']);
+    }
+    const map = options['filter'];
+    for (const key in map) {
+      if (!map[key]) {
+        throw new PineconeArgumentError('Filter cannot be empty');
```
austin-denoble, line 63: Could maybe add the `key` value to the error message here:
```suggestion
        throw new PineconeArgumentError(`Filter cannot be empty for ${key}`);
```

src/control/configureIndex.ts
austin-denoble: It seems like these `validator` functions are all defined as `async`. I don't think they should be, these functions shouldn't be doing any kind of network requesting or work which would require `async`.

src/utils/validateProperties.ts
```diff
@@ -0,0 +1,14 @@
+import { PineconeArgumentError } from '../errors';
+
+export const ValidateProperties = (item: any, validProperties: any[]) => {
```
austin-denoble, line 3: See my comment above about this: I think we need to do something like this:

```suggestion
function validateProperties<T>(item: T, validProperties: keysOf<T>)
```

Basically, we want to couple the type to the list of validation keys being passed in.

austin-denoble (commented):
Big thanks again for taking this one on, was a lot of work pulling all of that out. Overall this approach looks good to me but I've left some feedback. Primarily, let's avoid having these inline lists of hard-coded "keys" for the different type and try and find a better way of coupling the TS type to what we're providing for validation.

I think we can also clean up some of those pre-existing error messages from typebox which weren't formatted nicely.

src/control/configureIndex.ts
```diff
@@ -4,54 +4,50 @@ import {
...
-  ): Promise<IndexModel> => {
-    indexNameValidator(indexName);
-    patchRequestValidator(options);
+  ) => {
+    if (options) {
+      ValidateProperties(options, ['spec', 'deletionProtection']);
+    }
 
-    if (Object.keys(options).length === 0) {
+    if (!indexName || indexName.length === 0) {
```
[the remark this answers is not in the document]
aulorbe: Ah yes good catch! Harkening back to our t/f convo earlier :)

src/control/configureIndex.ts
```diff
@@ -4,54 +4,50 @@ import {
...
+    }
+    if (options.spec) {
+      if (options.spec.pod) {
+        ValidateProperties(options.spec.pod, ['replicas', 'podType']);
+      }
+      if (options.spec.pod && options.spec.pod.replicas) {
+        if (options.spec.pod.replicas <= 0) {
+          throw new PineconeArgumentError(
+            'The second argument to configureIndex had validation errors: property' +
+              " 'spec/properties/pod/properties/replicas' must be a positive integer."
```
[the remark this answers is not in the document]
aulorbe: oh sure, I can change!

src/control/createCollection.ts
```diff
@@ -3,27 +3,38 @@ import {
...
+      throw new PineconeArgumentError(
+        'The argument to createCollection must have required properties: name, source.'
+      );
+    }
+    if (!options.name || options.name.length === 0) {
+      throw new PineconeArgumentError(
+        'You must pass a non-empty string for `name` in order to create a collection.'
+      );
+    }
+    if (!options.source || options.source.length === 0) {
```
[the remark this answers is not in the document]
aulorbe: Ah yes these must've been old, apologies!

src/control/createCollection.ts
```diff
@@ -3,27 +3,38 @@ import {
...
+      ValidateProperties(options, ['name', 'source']);
+    }
+    if (!options || typeof options !== 'object') {
+      throw new PineconeArgumentError(
+        'You must pass a non-empty object with `name` and `source` fields in order to create a collection.'
+      );
+    }
+    if (!options.name && !options.source) {
+      throw new PineconeArgumentError(
+        'The argument to createCollection must have required properties: name, source.'
```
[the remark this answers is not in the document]
aulorbe: ah yes good call

src/control/deleteCollection.ts
aulorbe: It's prettier, I think! But I can try to shorten them

src/control/types.ts
aulorbe: Yeah that was a holdover from before.... I'm not sure why. I'd totally be pro removing it

src/control/createIndex.ts
aulorbe: Yeah, what I want to happen is: 
- Field names have backticks
- String values have single quotation marks

So in this instance, metric would look like ``metric``, and cosine would look like `'cosine'`

src/utils/validateProperties.ts
aulorbe: [update: we spoke offline and decided to simplify things -- we are just making custom constants to hold the property values & check against those]

austin-denoble (approved):
Nice work! I really appreciate all the iteration here around different ways we can handle validating the payloads for some of these requests. I'm excited to get both of these dependencies out of the codebase as well. 🚢

src/utils/validateProperties.ts
```diff
@@ -0,0 +1,20 @@
+import { PineconeArgumentError } from '../errors';
+import { CreateCollectionRequest } from '../pinecone-generated-ts-fetch/control';
+
+// Types and consts for property validation of generated Interfaces to ensure no unknown/invalid properties are passed, no req'd properties are missing
```
austin-denoble, line 4: Should this maybe live in the `src/control/createCollection.ts`? Seems like most of the specific constants for validation live in the respective implementation files.

aulorbe: Ah yes, great call, let's do it!

PR #250, opened 2024-08-15
Remove unnecessary crossFetch fallback

aulorbe:
## Problem

We made the decision a while back to [remove support for Node versions <18](https://github.com/pinecone-io/pinecone-ts-client/pull/239). Previously, in order to account for Node versions <18, w resorted to using a `ponyfill` called `crossFetch` for the `fetch` implementation, when no user-supplied implementation was found. Now that we no longer support Node versions <18, there's no need for us to fall back to `crossFetch`.

## Solution

Remove the `crossFetch` fallback.

## Type of Change

## Test Plan

CI passes.

---
  - https://app.asana.com/0/0/1208002049018704


src/utils/fetch.ts
```diff
@@ -10,9 +9,13 @@ export const getFetch = (config: PineconeConfiguration) => {
...
+    console.log(
+      'Failed to find any user-provided fetch implementation. Searching for global fetch implementation...'
+    );
     return global.fetch;
   } else {
-    // Use ponyfill as last resort
-    return crossFetch;
+    console.error(
+      'No global or user-provided fetch implementations found. Please supply a fetch implementation.'
+    );
```
austin-denoble, line 19: Do we want to actually throw here? I think this will just log an error to the console without interrupting control flow. If there's no available fetch implementation they wont be able to do anything with the client anyways, so maybe we should more forcefully exit here?

aulorbe: Good point, I'll turn it into a throw!

aulorbe: (ended up adding some unit tests to confirm everything, too)

austin-denoble:
I think there's some information in the `README` which could maybe be tweaked with this change: https://github.com/pinecone-io/pinecone-ts-client/blob/8bffd778b745865674a960907f9672c8dbbcdf12/README.md?plain=1#L900

aulorbe:
> I think there's some information in the `README` which could maybe be tweaked with this change:
> 
> https://github.com/pinecone-io/pinecone-ts-client/blob/8bffd778b745865674a960907f9672c8dbbcdf12/README.md?plain=1#L900

Thanks for finding this. Apologies for the oversight!

src/utils/fetch.ts
```diff
@@ -10,9 +10,13 @@ export const getFetch = (config: PineconeConfiguration) => {
...
+    console.log(
+      'Failed to find any user-provided fetch implementation. Using global fetch implementation.'
+    );
     return global.fetch;
   } else {
-    // Use ponyfill as last resort
-    return crossFetch;
+    throw new PineconeConfigurationError(
+      'No global or user-provided fetch implementations found. Please supply a fetch implementation.'
+    );
```
austin-denoble, line 20: I think throwing in this specific instance is fine. I'm trying to think of any unintended consequences from removing the generic fallback.

Are there possible instances where people were relying on `crossFetch` previously and now we're pulling it out? I don't think this would be likely given the support for >= Node 18.0.0, but was something I thought of while reviewing.

austin-denoble (approved):
LGTM overall.

src/utils/fetch.ts
aulorbe: I mean, technically they could be, but since removing support for Node <18, I think we relinquish responsibility for those users, no?

PR #252, opened 2024-08-21
Update client to use ESModules syntax

aulorbe:
## Problem

We recently [shipped a fix](https://github.com/pinecone-io/pinecone-ts-client/pull/249) to compatibility issues users were facing with our SDK when running in Edge runtimes.

While this seems to have fixed some problems for users running in Edge runtimes, another issue came to light while helping a community-forum poster: the way in which our Typescript files are transpiled (dictated by our `tsconfig` file) is in accordance with older Node versions, not more modern, popular Javascript syntax.

## More info
Our `tsconfig.json` file dictated that our Typescript files get transpiled into `es5` Javascript using the `CommonJS` module syntax. This is [usually reserved for ensuring users can run packages using old versions of Node](https://blog.logrocket.com/commonjs-vs-es-modules-node-js/) (e.g. <9). 

Since we made the decision to no longer support versions of Node <18, having our `tsconfig` file transpile our Typescript in this older syntax doesn't make a whole lot of sense.

To bring our transpilation process up to date with newer versions of Node, it follows that we should update our `"target"` to `"es2015"` and our `"module"` to `"ESNext"`. 

There are many key differences between the two (explained in the article linked above), but the most important difference is that they use different syntax for loading and exporting Javascript modules.

## How did this come to light?
There is a current, open issue on [our community forum](https://community.pinecone.io/t/error-using-node-js-sdk-on-edge/6337/11?u=audreylorb.pinecone) about using our SDK in an Edge runtime. While troubleshooting the Edge portion of this issue, it came to light that another problem this user was facing was related to how our Typescript files were being transpiled into Javascript.

When I looked into this user's stacktrace of the error, the highlighted lines were all lines that used the older, CommonJS module import syntax (`require`) instead of the newer ESModules import syntax (`import`). 

Updating our `tsconfig` file to use ESModules instead of CommonJS will ensure our transpiles files use the more modern Javascript syntax, which should, in turn, allow users to embed our client within their modern Javascript apps (running in Edge or Node runtimes). 

## Solution

Update all compilation-related files to transpile Typescript files using ESModules syntax. I've ensured all unit and integration tests pass locally.

## Notes
- As part of the general upgrade from using CommonJS to ESModules syntax, we also had to upgrade the ways in which some tests run + upgrade the syntax used by our jest config files. Changes include:
    - ESModules `import` and `export default` syntax replacing old `require` and `export.modules` syntax in our `jest.config....ts` files for Node and Edge
    - Add `"type": "module"` to our `package.json` files in both our root dir and our compilation dir we use for CI testing (mandatory for running ESModules)
    - Add conditional stmts to our `testing.yml` file, in the compilation section, so that Node `@types` are specific to the different TS versions we test in the matrix.

## Type of Change

## Test Plan

CI passes.

---
  - https://app.asana.com/0/0/1208096908844885

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "@pinecone-database/pinecone",
   "version": "3.0.1",
   "main": "dist/index.js",
+  "type": "module",
   "repository": {
     "type": "git",
     "url": "https://github.com/pinecone-io/pinecone-ts-client"
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "repl": "npm run build && node utils/replInit.ts",
     "test:integration:node": "TEST_ENV=node jest src/integration/ -c jest.config.integration-node.js --runInBand --bail",
     "test:integration:edge": "TEST_ENV=edge jest src/integration/ -c jest.config.integration-edge.js --runInBand --bail",
-    "test:integration:cleanup": "npm run build && node utils/cleanupResources.ts",
+    "test:integration:cleanup": "npm run build && npx utils/cleanupResources.ts",
     "test:unit": "jest src/"
   },
   "engines": {
@@ -45,6 +45,7 @@
     "jest-skipped-reporter": "^0.0.5",
     "prettier": "^2.8.8",
     "ts-jest": "^29.0.5",
+    "ts-node": "^10.9.2",
     "typedoc": "^0.24.8",
     "typescript": "^4.9.4"
   },
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "module": "commonjs",
+    "target": "es2015",
+    "module": "ESNext",
     "lib": ["es2015", "dom"],
     "declaration": true,
     "sourceMap": true,
@@ -12,7 +12,8 @@
     "forceConsistentCasingInFileNames": true,
     "types": ["node", "jest", "@edge-runtime/types"],
     "noImplicitAny": false,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "moduleResolution": "Node"
   },
   "include": ["src/**/*"],
   "exclude": ["node_modules"]
--- a/utils/cleanupResources.ts
+++ b/utils/cleanupResources.ts
@@ -1,5 +1,5 @@
 import dotenv from 'dotenv';
-import pinecone from '../dist/index.js'; // Explicitly include the '.js' extension
+import pinecone from '../dist/index.js';
 
 dotenv.config();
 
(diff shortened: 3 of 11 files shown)


tsconfig.json
```diff
@@ -12,7 +12,8 @@
     "forceConsistentCasingInFileNames": true,
     "types": ["node", "jest", "@edge-runtime/types"],
     "noImplicitAny": false,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "moduleResolution": "Node"
```
aulorbe, line 16: Ensures compatibility between CommonJS and ESModules syntax.

utils/cleanupResources.ts
```diff
@@ -9,21 +11,27 @@ for (const envVar of ['PINECONE_API_KEY']) {
...
-  for (const collection of collectionList.collections) {
-    console.log(`Deleting collection ${collection.name}`);
-    await p.deleteCollection(collection.name);
+  if (collectionList.collections) {
+    for (const collection of collectionList.collections) {
+      console.log(`Deleting collection ${collection.name}`);
+      await p.deleteCollection(collection.name);
+    }
+  } else {
+    console.log('No collections found to delete');
```
aulorbe, line 24: Added this if stmt (and the one below) b/c we were getting a TS error, since `collectionList.collections` could technically be undefined.

aulorbe: (We were likely _not_ getting these type errors when using CommonJS, since Typescript apparently interprets module boundaries and types more loosely in CommonJS vs ESModules. See [here](https://github.com/pinecone-io/pinecone-ts-client/pull/252/files#r1725714777), too.)

tsconfig.json
aulorbe: From what I'm reading , this now ensures that Typescript's module-resolution algorithm is `Node` (as opposed to `Classic`). 

`Node` vs `Classic` have different 'rule sets' Typescript assumes in order to provide correct `types` on imports at runtime (lifted from [here](https://www.typescriptlang.org/docs/handbook/modules/theory.html#esm-and-cjs-interoperability)).

package.json
```diff
@@ -23,7 +24,7 @@
     "repl": "npm run build && node utils/replInit.ts",
     "test:integration:node": "TEST_ENV=node jest src/integration/ -c jest.config.integration-node.js --runInBand --bail",
     "test:integration:edge": "TEST_ENV=edge jest src/integration/ -c jest.config.integration-edge.js --runInBand --bail",
-    "test:integration:cleanup": "npm run build && node utils/cleanupResources.ts",
+    "test:integration:cleanup": "npm run build && npx utils/cleanupResources.ts",
```
aulorbe, line 27: changed to using `npx` here since with our new setup, we cannot run `.ts` files directly via `node`

aulorbe:
Note: we are deprioritizing this work, as updating the client to work with ESM syntax is a larger project than we want to take on right now.

PR #253, opened 2024-08-22
Update GH workflow files to use Node 20.x

aulorbe:
## Problem

We were getting warnings about using a deprecated version of a Github workflow (`cache@3`).

We were also getting warnings about using an input called `node_version` that was invalid. 

## Solution

Update the workflow we use from `@3` to `@4`. Additionally, add `node_version` into `inputs` section.

Additional housekeeping change: remove `crossFetch` from `package.json`, it was just there as a holdover from when we used `crossFetch`, but [we have removed it](https://github.com/pinecone-io/pinecone-ts-client/pull/250).

## Type of Change

## Test Plan

CI passes.

---
  - https://app.asana.com/0/0/1208096908844879

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "/dist"
   ],
   "dependencies": {
-    "cross-fetch": "^3.1.5",
     "encoding": "^0.1.13"
   }
 }
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -4,6 +4,9 @@ inputs:
   pinecone_api_key:
     description: 'API key'
     required: true
+  node_version:
+    description: 'Node.js version to setup'
+    required: true
 runs:
   using: 'composite'
   steps:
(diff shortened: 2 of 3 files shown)


aulorbe:
Examples of [warnings in CI we were getting](https://github.com/pinecone-io/pinecone-ts-client/actions/runs/10515054459):

Deprecated Node version one:
Screenshot 2024-08-22 at 2 22 20 PM

`node_version` one:
Screenshot 2024-08-22 at 2 22 27 PM

package.json
```diff
@@ -52,7 +52,6 @@
     "/dist"
   ],
   "dependencies": {
-    "cross-fetch": "^3.1.5",
```
austin-denoble, line 55: Oh, did we not actually remove this in the previous PR? https://github.com/pinecone-io/pinecone-ts-client/pull/250/files

austin-denoble (approved):
LGTM, thanks for fixing!

package.json
aulorbe: lol nope

PR #256, opened 2024-08-27
Update README to no longer mention runtime checks

aulorbe:
## Problem

Since we removed runtime checks, we need to remove the part in our README about the `PINECONE_DISABLE_RUNTIME_VALIDATIONS` env var (missed 1st time around): https://github.com/pinecone-io/pinecone-ts-client/pull/249/files#diff-a54261af011358e447a9f3ef717ee8e5d5bb4258f8d95e246b39cf3666244f1fL275

Screenshot 2024-08-26 at 5 23 35 PM

## Solution

Remove it :) 

## Type of Change

### DIFF
--- a/README.md
+++ b/README.md
@@ -966,8 +966,5 @@ generateQueryEmbeddings().then((embeddingsResponse) => {
 
 ## Productionizing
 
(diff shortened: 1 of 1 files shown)


austin-denoble (approved):
🚢

PR #278, opened 2024-08-29
Add final, working Vercel app e2e Edge GH workflow

aulorbe:
## Problem

Setting the necessary Pinecone-related environment variables in the Vercel UI enabled us to finalize the e2e Edge GH workflow pipeline. Everything now deploys and is workable in the `preview` environment (all we need for right now). 

Final working preview: https://vercel.com/pinecone-io/pinecone-rag-demo/G6hqB1w2fdNyLbHhdjqYvx7pgVzo

## Solution

Pass Pinecone-related env vars needed by app through Vercel UI/dashboard.

## Type of Change

---
  - https://app.asana.com/0/0/1208153851794098

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,11 @@
     "es2021": true,
     "node": true
   },
-  "ignorePatterns": ["dist", "src/pinecone-generated-ts-fetch", "pinecone-rag-demo"],
+  "ignorePatterns": [
+    "dist",
+    "src/pinecone-generated-ts-fetch",
+    "pinecone-rag-demo"
+  ],
   "extends": [
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,7 @@
     "es2021": true,
     "node": true
   },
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -4,7 +4,7 @@
     "es2021": true,
     "node": true
   },
-  "ignorePatterns": ["dist", "src/pinecone-generated-ts-fetch"],
```
aulorbe, line 7: Needed to add this here bc for some reason Vercel was reading the client's linting config instead of the cloned repo's.

aulorbe: (eventually, when we use what we learned here to build our own, lightweight version of a Vercel app for our Edge testing, we'll remove this, since we won't be cloning someone else's Vercel project)

austin-denoble: I think this is fine for now also, we don't need to worry about their linting if we're just cloning things.

austin-denoble (approved):
So the values are set in the Vercel console rather than passed via the workflow?

That works for now, I wonder if it had something to do with those errors you were seeing around variable naming. We can probably troubleshoot more deeply when working on the lightweight test harness.

PR #279, opened 2024-08-30
Update cleanupResources workflow to check for deletionProtection

aulorbe:
## Problem

Currently, the `cleanupResources` script [fails](https://github.com/pinecone-io/pinecone-ts-client/actions/runs/10623637432/job/29452705669?pr=278) to delete some indexes that spin up during our integration test suite because some of these indexes have `deletionProtection` `enabled`. 

## Solution

Check for `deletionProtection` being `enabled`; if it is, set it to `disabled`.

## Type of Change

## Test Plan

CI passes

---
  - https://app.asana.com/0/0/1208192096733434
  - https://app.asana.com/0/0/1208186698556887

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,7 @@
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-inferrable-types": "off",
-    "import/no-cycle": "error"
+    "import/no-cycle": "error",
+    "@typescript-eslint/no-var-requires": "off" // Add this line
   }
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,6 @@
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-inferrable-types": "off",
-    "import/no-cycle": "error",
-    "@typescript-eslint/no-var-requires": "off"
+    "import/no-cycle": "error"
   }
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,6 +25,6 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-inferrable-types": "off",
     "import/no-cycle": "error",
-    "@typescript-eslint/no-var-requires": "off" // Add this line
+    "@typescript-eslint/no-var-requires": "off"
   }
 }
--- a/utils/cleanupResources.ts
+++ b/utils/cleanupResources.ts
@@ -1,4 +1,7 @@
-var dotenv = require('dotenv');
+import dotenv from 'dotenv';
+
+import pinecone from '../dist';
+
 dotenv.config();
 
 for (const envVar of ['PINECONE_API_KEY']) {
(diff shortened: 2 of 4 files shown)


utils/cleanupResources.ts
```diff
@@ -1,4 +1,7 @@
-var dotenv = require('dotenv');
+import dotenv from 'dotenv';
```
aulorbe, line 1: Change to be up to date w/modern JS syntax

utils/cleanupResources.ts
```diff
@@ -8,23 +11,27 @@ for (const envVar of ['PINECONE_API_KEY']) {
...
-var pinecone = require('../dist');
-
 (async () => {
   const p = new pinecone.Pinecone();
 
   const collectionList = await p.listCollections();
-  for (const collection of collectionList.collections) {
-    console.log(`Deleting collection ${collection.name}`);
-    await p.deleteCollection(collection.name);
+  if (collectionList.collections) {
```
aulorbe, line 18: Same here -- this was throwing errors in my IDE because `collectionList.collections` could be undefined.

.eslintrc.json
```diff
@@ -24,6 +24,7 @@
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-inferrable-types": "off",
-    "import/no-cycle": "error"
+    "import/no-cycle": "error",
+    "@typescript-eslint/no-var-requires": "off"
```
aulorbe, line 28: Long story w/this one, but essentially I added this because: 
- Our `eslint` config is set up to expect `es6` modules (uses `import` for imports) on the line `"sourceType": "module" `. 
- Our `tsconfig` file, though, tells the compiler to use CJS syntax (uses `require` for imports)
- So, linting complains about the `require` statements in our `utils/` files without the addition of this rule to our `eslint` config 

We should make whatever syntax we want to use consistent throughout the repo to avoid confusing users at some point, but for now this is the most minimal change that allows me to bypass linting issues.

aulorbe: Deeper dive:

Q: Why doesn't my IDE complain about the files in dirs such as `src/` using `import` instead of `require`, if `tsconfig` says we're supposed to be writing CJS?
- This is because we also have the line "esModuleInterop": true` in our `tsconfig` file. This allows us to use ESM syntax (`import`) in our Typescript files, even though the underlying module system is CJS. 

Q: Why don't you just change the imports in the `utils/` files to use `import` (ESM) instead of `require` (CJS) then?
- Wwhen I tried [this](https://github.com/pinecone-io/pinecone-ts-client/actions/runs/10636403435/job/29489510906), I ran into the following error: 
Screenshot 2024-08-30 at 11 40 55 AM
- I think this we run the files in `utils/` with the `node` command, and `node` does not understand `typescript` or `es6` (ESM) module syntax. By default, `node` expects `CJS` syntax.

utils/cleanupResources.ts
```diff
@@ -8,23 +11,33 @@ for (const envVar of ['PINECONE_API_KEY']) {
...
+      console.log(`Deleting collection ${collection.name}`);
+      await p.deleteCollection(collection.name);
+    }
   }
 
   const response = await p.listIndexes();
-  for (const index of response.indexes) {
-    console.log(`Deleting index ${index.name}`);
-    await p.deleteIndex(index.name);
+  if (response.indexes) {
```
aulorbe, line 26: (Same here wrt `undefined`)

utils/cleanupResources.ts
```diff
@@ -1,4 +1,7 @@
-var dotenv = require('dotenv');
+const dotenv = require('dotenv');
```
aulorbe, line 1: Since this item doesn't change, I switched it to `const` on my IDE's advice :)

.eslintrc.json
austin-denoble: Interesting, thanks for doing a deep dive on this. I see the IDE issues as well, but since we specifically lint `src/` I think that's why this isn't generally an issue in CI and whatever.

I'd personally prefer something more specific more specific than disabling for the whole repo. I think we should probably look at cleaning some of these utility files up a bit.

austin-denoble (approved):
LGTM, thanks for cleaning this up!

.eslintrc.json
aulorbe: [ignore this -- made a ticket to address in a separate PR: https://app.asana.com/0/1203260648987893/1208186698556902/f]

PR #284, opened 2024-09-19
Add rerank endpoint

aulorbe:
## Problem

This PR adds the new [rerank endpoint](https://docs.pinecone.io/guides/inference/rerank) into the TS client.

Note, README reflects addition.

## Type of Change

## Test Plan

CI passes. Added new unit & integration tests.

---
  - https://app.asana.com/0/0/1208242198563722

### DIFF
--- a/src/inference/__tests__/inference.test.ts
+++ b/src/inference/__tests__/inference.test.ts
@@ -2,6 +2,7 @@ import { Inference } from '../inference';
 import type { PineconeConfiguration } from '../../data';
 import { inferenceOperationsBuilder } from '../inferenceOperationsBuilder';
 import { PineconeArgumentError } from '../../errors';
+import { Rerank200Response } from '../../pinecone-generated-ts-fetch/inference';
 
 describe('Inference Class: _formatInputs', () => {
   let inference: Inference;
@@ -89,12 +90,13 @@ describe('Inference Class: rerank', () => {
     }
   });
 
-  test('Throws error if list of dicts is passed for docs, but does not contain `text` key', async () => {
+  test('Throws error if list of objs is passed for docs, but does not contain `text` key', async () => {
     const rerankingModel = 'test-model';
     const myQuery = 'test-query';
     const myDocuments = [{ id: '1' }, { id: '2' }];
     const expectedError = new PineconeArgumentError(
-      '`documents` can only be a list of strings or a list of dictionaries with at least a `text` key, followed by a string value'
+      '`documents` can only be a list of strings or a list of objects with at least a `text` key, followed by a' +
+        ' string value'
     );
     try {
       await inference.rerank(rerankingModel, myQuery, myDocuments);
@@ -103,14 +105,17 @@ describe('Inference Class: rerank', () => {
     }
   });
 
-  test('Confirm list of strings as docs is converted to list of dicts with `text` key', async () => {
+  test('Confirm list of strings as docs is converted to list of objects with `text` key', async () => {
     const rerankingModel = 'test-model';
     const myQuery = 'test-query';
     const myDocuments = ['doc1', 'doc2'];
     const expectedDocuments = [{ text: 'doc1' }, { text: 'doc2' }];
     const rerank = jest.spyOn(inference._inferenceApi, 'rerank');
-    // @ts-ignore
-    rerank.mockResolvedValue({ rerankResponse: {} });
+    rerank.mockResolvedValue({
+      model: 'some-model',
+      data: [{}],
+      usage: { rerankUnits: 1 },
+    } as Rerank200Response);
     await inference.rerank(rerankingModel, myQuery, myDocuments);
 
     const expectedReq = {
--- a/src/inference/__tests__/inference.test.ts
+++ b/src/inference/__tests__/inference.test.ts
@@ -2,7 +2,7 @@ import { Inference } from '../inference';
 import type { PineconeConfiguration } from '../../data';
 import { inferenceOperationsBuilder } from '../inferenceOperationsBuilder';
 import { PineconeArgumentError } from '../../errors';
-import { Rerank200Response } from '../../pinecone-generated-ts-fetch/inference';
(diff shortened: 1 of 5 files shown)


src/inference/inference.ts
```diff
@@ -3,8 +3,18 @@ import {
   EmbedRequestInputsInner,
   EmbedRequestParameters,
   InferenceApi,
+  Rerank200Response as RerankResponse,
 } from '../pinecone-generated-ts-fetch/inference';
 import { EmbeddingsList } from '../models';
+import { PineconeArgumentError } from '../errors';
+import { prerelease } from '../utils/prerelease';
+
+export interface RerankOptions {
```
austin-denoble, line 12: We should make sure that this type and the generated types that are used for inference are exported from the top level of the package so people have access to them: https://github.com/pinecone-io/pinecone-ts-client/blob/73d98b8aaffd93dafadab8902f6728b153ae25a9/src/index.ts#L45-L63

I'm guessing there might be a bunch of embedding / rerank stuff that we're using in these functions and maybe also need to be exposed.

src/inference/__tests__/inference.test.ts
```diff
@@ -65,3 +66,146 @@ describe('Inference Class: embed', () => {
...
+      'You must pass at least one document to rerank'
+    );
+    try {
+      await inference.rerank(rerankingModel, myQuery, []);
+    } catch (error) {
+      expect(error).toEqual(expectedError);
+    }
+  });
+
+  test('Throws error if list of dicts is passed for docs, but does not contain `text` key', async () => {
```
austin-denoble, line 92: nit: They're objects not dicts in javascript, technically. Sorry I know that's pedantic. I would just replace all dict/dictionary references to objects.

src/inference/__tests__/inference.test.ts
```diff
@@ -65,3 +66,146 @@ describe('Inference Class: embed', () => {
...
+  });
+
+  test('Confirm list of strings as docs is converted to list of dicts with `text` key', async () => {
+    const rerankingModel = 'test-model';
+    const myQuery = 'test-query';
+    const myDocuments = ['doc1', 'doc2'];
+    const expectedDocuments = [{ text: 'doc1' }, { text: 'doc2' }];
+    const rerank = jest.spyOn(inference._inferenceApi, 'rerank');
+    // @ts-ignore
+    rerank.mockResolvedValue({ rerankResponse: {} });
```
austin-denoble, line 113: Do we need the `@ts-ignore` here because of the empty response? Can we just create a bogus `RerankResponse` inline? I'd prefer to avoid using `@ts-ignore` if possible.

src/inference/inference.ts
```diff
@@ -3,8 +3,18 @@ import {
   EmbedRequestInputsInner,
   EmbedRequestParameters,
   InferenceApi,
+  Rerank200Response as RerankResponse,
```
austin-denoble, line 6: This naming is kind of unfortunate, the type in the generated code. Should we maybe look into fixing the spec upstream to make this a little cleaner? I'd need to look at it but I think there's a way we've resolved some of this naming stuff by tweaking the spec. I think there may be a model that needs the x-component tag or whatever.

Just a thought.

src/inference/inference.ts
```diff
@@ -47,4 +57,136 @@ export class Inference {
...
+  /** Rerank documents against a query with a reranking model. Each document is ranked in descending relevance order
+   *  against the query provided.
+   *
+   *  Note: by default, the ['text'] field of each document is used for ranking; you can overwrite this default
+   *  behavior by passing an {@link RerankOptions} `options` object specifying 1+ other fields.
+   *
+   *  @example
+   *  ```typescript
+   *  import { Pinecone } from '@pinecone-database/pinecone';
+   *  const pc = new Pinecone();
```
austin-denoble, line 70: nit: Can we add a line break here just to visually break this up a bit?

src/inference/inference.ts
```diff
@@ -47,4 +57,136 @@ export class Inference {
...
+      returnDocuments = true,
+      rankFields = ['text'],
+      parameters = {},
+    } = options;
+
+    // Allow documents to be passed a list of strings, or a list of dicts w/at least a `text` key:
+    let newDocuments: Array<{ [key: string]: string }> = [];
+    if (typeof documents[0] === 'object' && !('text' in documents[0])) {
+      throw new PineconeArgumentError(
+        '`documents` can only be a list of strings or a list of dictionaries with at least a `text` key, followed by a string value'
```
austin-denoble, line 155: Would definitely swap dictionaries -> objects here.

austin-denoble (approved):
Overall looks good, I did leave some nit feedback around verbiage and a few other questions. Nice work. 🚢

src/inference/inference.ts
aulorbe: So, for inputs, the users shouldn't need the exported types or interfaces, since we wrap those in more ux-friendly stuff. But I've now added the rerank response types to this file, good call. (The response type `EmbeddingsList` for /embed is wrapped, and we export that.

src/inference/__tests__/inference.test.ts
aulorbe: Will do!

src/inference/__tests__/inference.test.ts
aulorbe: Will do!

src/inference/inference.ts
aulorbe: yes!

src/inference/inference.ts
aulorbe: Good thinking -- I've opened a PR here https://github.com/pinecone-io/apis/pull/147

src/inference/inference.ts
austin-denoble: Yup, thanks!

austin-denoble (approved):

PR #287, opened 2024-09-23
Reorganize RC branch in anticipate of Oct release

aulorbe:
## Problem

We now have two different prefixes that append to the data plane API : `bulk` and `vectors`. As such, it's time to reorganize our repo! 

## Solution

The `data/` dir now is organized into 2 subdirs: `bulk/` and `vectors/`.  Tests share the same organization.

`data/index.ts` is still at the `data/` dir level (as is `indexHostSingleton.ts`), since that's the entry point for both prefixes to reach the data plane (i.e. the `Index` class); tests reflect this as well.

## Type of Change

## Test Plan

CI passes.


austin-denoble (approved):
LGTM. I did want to call out that to some degree we've tried to follow the naming pattern of "record" instead of "vector" in the TypeScript codebase: https://github.com/pinecone-io/pinecone-ts-client?tab=readme-ov-file#upsert-records

Do you think it would make sense to go with `/records/` instead of `/vectors/` because of this?

aulorbe:
> https://github.com/pinecone-io/pinecone-ts-client?tab=readme-ov-file#upsert-records

(We aligned as a team offline about this, and we will stick with `vector`)

PR #288, opened 2024-09-25
Change language around records --> vectors where appropriate

aulorbe:
## Problem

We believe it's more accurate to refer to dense-vector objects as "vectors" (vs "records") when we are talking about vectors we'd like to upsert into an index, etc. This is because they are primarily (first and foremost) _vectors_, with IDs, metadata, etc.

## Solution

Change instances of "record(s)" to "vector(s)" where appropriate.

Note: I opted _not_ to change instances of "record(s)" in cases where we are talking about the entire object as a whole (as in, "Update a record", since you could be updating the metadata, the vector values, etc.).

## Type of Change

---
  - https://app.asana.com/0/0/1208381566824289

### DIFF
--- a/README.md
+++ b/README.md
@@ -653,7 +653,7 @@ See [Use namespaces](https://docs.pinecone.io/guides/indexes/use-namespaces) for
 
 ### Upsert vectors
 
-Pinecone expects vectors inserted into indexes to have the following form:
+Pinecone expects records inserted into indexes to have the following form:
 
 ```typescript
 type PineconeRecord = {
--- a/README.md
+++ b/README.md
@@ -141,7 +141,9 @@ indexes().then((response) => {
 
 #### Create a serverless index with minimal configuration
 
-At a minimum, to create a serverless index you must specify a `name`, `dimension`, and `spec`. The `dimension` indicates the size of the records you intend to store in the index. For example, if your intention was to store and query embeddings generated with OpenAI's [textembedding-ada-002](https://platform.openai.com/docs/guides/embeddings/second-generation-models) model, you would need to create an index with dimension `1536` to match the output of that model.
+At a minimum, to create a serverless index you must specify a `name`, `dimension`, and `spec`. The `dimension`
+indicates the size of the vectors you intend to store in the index. For example, if your intention was to store and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -688,13 +690,13 @@ const records = [
...
 // Upsert the data into your index
-await index.upsert(records);
+await index.upsert(vectors);
 ```
 
 ### Seeing index statistics
 
-When experimenting with data operations, it's sometimes helpful to know how many records are stored in each namespace. In that case,
-target the index and use the `describeIndexStats()` command.
+When experimenting with data operations, it's sometimes helpful to know how many records/vectors are stored in each
```
aulorbe, line 698: made this "records/vectors" just b/c I could see using both here

README.md
```diff
@@ -800,7 +802,7 @@ const index = pc.index('my-index');
 const results = await index.query({ topK: 10, id: '1' });
 ```
 
-#### Hybrid search with sparseVector
+#### Hybrid search with sparse vectors
```
aulorbe, line 805: Updating this example while I'm here in this PR -- unrelated to language, but we were using a pretty outdated image search example (really, that's "multi-modal" search) that I didn't find intuitive.

README.md
```diff
@@ -649,9 +651,9 @@ await index.fetch(['1']);
 
 See [Use namespaces](https://docs.pinecone.io/guides/indexes/use-namespaces) for more information.
 
-### Upsert records
+### Upsert vectors
 
-Pinecone expects records inserted into indexes to have the following form:
+Pinecone expects vectors inserted into indexes to have the following form:
```
jhamon, line 656: Records is the correct terminology for vector values + id + metadata which is what you're dealing with here. https://docs.pinecone.io/guides/get-started/key-concepts#record

austin-denoble (approved):
Nice, I like the approach here. Cleaning up the language in some of the documentation and leaving the "record" terminology elsewhere makes sense given how we used it all over the codebase, and it still feels accurate.

README.md
aulorbe: ah okay, I'll change that one back then, thanks!

PR #291, opened 2024-10-03
Centralize test indexes for integration tests

aulorbe:
## Problem

Our integration tests are currently very flakey (meaning they fail in an non-deterministic manner depending largely on fluctuations in latency on backend operations). 

## Solution

Centralize the creation and deletion of indexes for integration tests; cleanup the tests where I can.

**The tests that now use the centralized indexes:** 
**- Control:**
    - `describeIndex.test.ts`
    - `listIndexes.test.ts`
**- Data:**
    -  `fetch.test.ts`
    - `list.test.ts`
    - `query.test.ts`

**The tests that still have to create/delete their own indexes because their operations are difficult to undo:**
**- Control:**
    - `configureIndex.test.ts`
    - `createIndex.test.ts`
**- Data:**
    - `delete.test.ts`
    -  `upsertAndUpdate.test.ts`

### New jobs & files
- I added a job in `testing.yml` called `setup` that calls a file called `setup.ts` (`src/integration/setup.ts`)
- I added a job in `testing.yml` called `teardown`  that calls a file called `teardown.ts` (`src/integration/teardown.ts`)
- `setup.ts` sets up (creates and seeds) a shared serverless index (only serverless for now b/c we don't have pod-based data plane tests). **The index name is randomized, so simultaneous jobs can occur using the same Pinecone API key** 😄 .
    - I have put a `todo` in this file to refactor it; right now, it's quite rudimentary in the way it loops and checks conditions; you'll notice other `todo`s in other files that point to where we need more tests, etc. I'd like to keep these in there for now, if that's okay. 
- `teardown.ts` deletes this index
- I found some tests that were not actually testing what they intended to, so I have marked those as skipped for now (e.g. `createIndex.test.ts > `test('insufficient quota')`)
- I added a file called `integrationRunner.js`, which is a file for running integration tests locally with a new command in `package.json` that you can execute by running `npm run test:integration-local:node`. This will do the setup, tests, and teardown all in 1 go for you locally.

### Bug fixes
- Last week, @austin-denoble and I noticed a bug in `waitUntilReady` where the `status` sometimes evaluated to `Ready` without the index actually being ready. So, this PR includes a fix to that function that also checks for `state.ready` being `true` _and_ adds a value (that it also checks for) to the `IndexModelStatusStateEnum` that accounts for an index's state being `Upgrading`, which accounted for a lot of flakiness in the configureIndex tests.

## Notes
- The goal of this PR is to _decrease_ flakiness
- I removed one of our Bun versions to try to get the time down a bit; we might want to experiment with upping the `max-parallel` field too in `testing.yml` (although I tried removing it altogether and we had failures)
- I _removed_ the job in `testing.yml` that calls `utils/cleanupResources.ts`, since I replaced it with my `Teardown` job. This file (`cleanupResources`) is still called via cron, though, by `testing-cleanup.yml`
- I decided _not_ to use [jest's `globalSetup` and `globalTeardown` capabilities](https://jestjs.io/docs/configuration#globalsetup-string) because the "global" scope is per option in our test matrix. Creating a custom setup and teardown job instead allowed me to create a single test index we can use across all options in our matrix, which is more efficient.

## Type of Change

## Test Plan

CI passes consistently

---
  - https://app.asana.com/0/0/1208439494339500
  - https://app.asana.com/0/0/1207545518662985


.github/workflows/testing.yml
```diff
@@ -4,7 +4,24 @@ on:
   workflow_call: {}
 
 jobs:
+  setup:
```
aulorbe, line 7: New `setup` job

.github/workflows/testing.yml
```diff
@@ -4,7 +4,24 @@ on:
...
+        run: npm ci
+
+      - name: Run setup script
+        env:
+          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
+        run: |
+          npx ts-node ./src/integration/setup.ts
+
   unit-tests:
+    needs: setup
```
aulorbe, line 24: `needs` means `setup` has to run before these jobs can start

src/integration/data/delete.test.ts
```diff
@@ -1,164 +1,86 @@
...
-        fail(
-          'Did not find expected records. Fetch result was ' +
-            JSON.stringify(results)
-        );
-      }
-    };
+// todo: deleting non-existent records
+// todo: delete all records in namespace
+// todo: delete namespace
+// todo: delete index (?)
```
aulorbe, line 50: Just outlining tests I'd like to add

aulorbe:
(Recent CI failures are due to super flakey upsert-and-update test. Have removed for now. No way to win with that test despite adding a big sleep. Will revisit later.)

aulorbe:
> (Recent CI failures are due to super flakey upsert-and-update test. Have removed for now. No way to win with that test despite adding a big sleep. Will revisit later.)

Okay. @austin-denoble and I decided that we will assert the client calls the correct endpoint w/the correct params, but _not_ assert things like updates, etc. are actually reflected later. This eliminates a ton of waiting time + works towards separation of concerns, since if updates (e.g.) don't work, that is ultimately a bug with the backend, not the client.

src/pinecone-generated-ts-fetch/control/models/IndexModelStatus.ts
```diff
@@ -45,6 +45,7 @@ export const IndexModelStatusStateEnum = {
   ScalingDownPodSize: 'ScalingDownPodSize',
   Terminating: 'Terminating',
   Ready: 'Ready',
+  Upgrading: 'Upgrading',
```
aulorbe, line 48: see new `Bug fixes` section in PR description for more info

src/integration/test-helpers.ts
```diff
@@ -102,11 +105,25 @@ export const sleep = async (ms) => {
 export const waitUntilReady = async (indexName: string) => {
```
aulorbe, line 105: Refactored this. See new `Bug fixes` section in PR description for more info

aulorbe:
We have had 3 successful* runs in a row. I am declaring the flakiness (that we control) surpassed!

*There is [a bug](https://app.asana.com/0/1204819992273155/1207992392793971/f) with `list` where sometimes a pagination token is not included in the response. Not counting this as a 'flake' on our end.

src/integration/setup.ts
```diff
@@ -0,0 +1,108 @@
+import { IndexList, Pinecone } from '../index';
+import {
+  diffPrefix,
+  generateRecords,
+  globalNamespaceOne,
+  prefix,
+  serverlessIndexName,
```
austin-denoble, line 7: I mentioned this in another comment, but I really don't think we should be using a hard-coded value for the index name, I think it should be a randomized and specific to a single test run, so across all files.

src/pinecone-generated-ts-fetch/control/models/IndexModelStatus.ts
austin-denoble: This model is from generated code, so we shouldn't be updating it manually. This corresponds to this in the spec: https://github.com/pinecone-io/apis/blob/0098ac4ff262286f371bb8b6803549a3446e5065/src/release/db/control/resources/indexes/models/IndexModel.yaml#L58-L66

I don't think we actually represent it in the spec at all, but it should probably be fixed upstream if we need to.

austin-denoble (commented):
> I decided not to use [jest's globalSetup and globalTeardown capabilities](https://jestjs.io/docs/configuration#globalsetup-string) because they interfered with how our integration tests run in parallel in CI -- the "global"-ness of these jest fixtures would be per option in our matrix, so even though they would be "global" to all test files, they would not be global across all CI runs (node, bun, edge, etc.), so collisions happened.

When I initially was thinking about updating things I was thinking "global" in terms of all files - so a single "test" run would be per option in the matrix. 

Instead of having a hard-coded "serverlessIndexName" that gets inherently shared, I'd prefer if each run of integration tests had an index name that was randomized, and then shared across all test files in that run. 

I'm not sure what issues you were seeing when using the global setup / teardown in Jest. The other client integration test suites tend to create randomized names for indexes and resources per run, so things are shared globally across integration tests / files, but not all possible integration tests that might be run in CI when factoring in a matrix.

src/integration/data/list.test.ts
```diff
@@ -1,89 +1,56 @@
...
+});
 
-    await ns.upsert(recordsToUpsert);
-    await waitUntilRecordsReady(ns, namespace, upsertedIds);
-    recordIds.concat(upsertedIds);
+describe('listPaginated, serverless index', () => {
+  test('test listPaginated with no arguments', async () => {
+    const listResults = await serverlessIndex.listPaginated();
+    expect(listResults).toBeDefined();
+    // expect(listResults.pagination).toBeDefined(); todo: re-enable this once pagination bug is fixed (https://app.asana.com/0/1204819992273155/1207992392793971/f)
```
aulorbe, line 24: Commenting out pagination assertions until bug is fixed (Asana ticket in comment)

package.json
```diff
@@ -23,6 +23,7 @@
     "repl": "npm run build && node utils/replInit.ts",
     "test:integration:node": "TEST_ENV=node jest src/integration/ -c jest.config.integration-node.js --runInBand --bail",
     "test:integration:edge": "TEST_ENV=edge jest src/integration/ -c jest.config.integration-edge.js --runInBand --bail",
+    "test:integration-local:node": "node src/integration/integrationRunner.js",
```
aulorbe, line 26: For local runs

src/integration/control/configureIndex.test.ts
```diff
@@ -42,12 +50,10 @@ describe('configure index', () => {
   });
 
   afterAll(async () => {
-    // wait until indexes are done upgrading before deleting
-    await waitUntilReady(podIndexName);
-    await waitUntilReady(serverlessIndexName);
-
-    await pinecone.deleteIndex(podIndexName);
-    await pinecone.deleteIndex(serverlessIndexName);
+    // Note: using retryDeletes instead of waitUntilReady due to backend bug where index status is ready, but index
```
aulorbe, line 53: Call out; will make a PR upstream to fix this

aulorbe:
> > I decided not to use [jest's globalSetup and globalTeardown capabilities](https://jestjs.io/docs/configuration#globalsetup-string) because they interfered with how our integration tests run in parallel in CI -- the "global"-ness of these jest fixtures would be per option in our matrix, so even though they would be "global" to all test files, they would not be global across all CI runs (node, bun, edge, etc.), so collisions happened.
> 
> When I initially was thinking about updating things I was thinking "global" in terms of all files - so a single "test" run would be per option in the matrix.
> 
> Instead of having a hard-coded "serverlessIndexName" that gets inherently shared, I'd prefer if each run of integration tests had an index name that was randomized, and then shared across all test files in that run.
> 
> I'm not sure what issues you were seeing when using the global setup / teardown in Jest. The other client integration test suites tend to create randomized names for indexes and resources per run, so things are shared globally across integration tests / files, but not all possible integration tests that might be run in CI when factoring in a matrix.

Note: we spoke about this offline. We are sticking w/the current approach of having setup and teardown jobs, since it's more efficient, _but_ we are randomizing the index names to allow simultaneous runs by diff users on the same Pinecone account 🎉

src/integration/setup.ts
```diff
@@ -0,0 +1,105 @@
...
+  diffPrefix,
+  generateRecords,
+  globalNamespaceOne,
+  prefix,
+  sleep,
+} from './test-helpers';
+
+// todo: refactor to make conditions & loops more efficient
+
+const setup = async () => {
```
austin-denoble, line 12: nit: I'd maybe add comments to the setup and teardown functions that call out the need for passing `SERVERLESS_INDEX_NAME` through environment variables, just so it's clear.

package.json
austin-denoble: Can we update the existing `test:integration:node` and `test:integration:edge` commands to handle calling `integrationRunner` rather than creating new calls that also run the tests with the proper setup and teardown?

.github/workflows/testing.yml
```diff
@@ -4,7 +4,30 @@ on:
...
+      - name: Install dependencies
+        id: step2
+        run: npm ci
+
+      - name: Run setup script
+        id: step3
+        env:
+          PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
+        run: |
+          SERVERLESS_INDEX_NAME=$(npx ts-node ./src/integration/setup.ts | grep "SERVERLESS_INDEX_NAME=" | cut -d'=' -f2)
```
austin-denoble, line 26: Ah I see, we don't use the `integrationRunner` here, only for local runs.

austin-denoble (approved):
Really great work on this, Audrey! The tests run super fast with this universally shared setup, and they seem quicker and more robust overall.

The only feedback I really have is that it would be nice to smooth out the local runs portion of things, or at least clean up the `package.json` file so there aren't multiple ways of running integration tests locally. Otherwise this looks ready to go. 🚢

.github/workflows/testing.yml
aulorbe: yaya :)

src/integration/setup.ts
```diff
@@ -0,0 +1,106 @@
...
+      //   upsert records into namespace
+      await pc
+        .index(randomIndexName)
+        .namespace(globalNamespaceOne)
+        .upsert(allRecords);
+
+      await sleep(10000);
+    }
+  }
+  // Capture output in GITHUB_OUTPUT env var when run in CI; necessary to pass across tests
```
aulorbe, line 102: @austin-denoble re: [this comment of yours](https://github.com/pinecone-io/pinecone-ts-client/pull/291/files#r1796171508), is this comment okay?

austin-denoble (approved):

PR #292, opened 2024-10-07
Update rerank section of README

aulorbe:
## Summary

After speaking with Silas about the `score` returned by the `/rerank` endpoint, I wanted to clarify its meaning for users a bit more.

## Type of Change

---
  - https://app.asana.com/0/0/1208413594221574

### DIFF
--- a/README.md
+++ b/README.md
@@ -1083,9 +1083,9 @@ generateQueryEmbeddings().then((embeddingsResponse) => {
 
 Rerank documents in descending relevance-order against a query.
 
-**Note:** The relevance `score` returned by the `/rerank` endpoint is _absolute_. This means that it is possible to
-compare `scores` across different calls to the `/rerank` endpoint. Each `score` is deterministic and represents how
-closely matched each specific item is to the query string.
+**Note:** The `score` represents the absolute measure of relevance of a given query and passage pair. Normalized
+between [0, 1], the `score` represents how closely relevant a specific item and query are, with scores closer to 1
+indicating higher relevance.
 
 ```typescript
 import { Pinecone } from '@pinecone-database/pinecone';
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1083,6 +1083,10 @@ generateQueryEmbeddings().then((embeddingsResponse) => {
 
 Rerank documents in descending relevance-order against a query.
 
+**Note:** The relevance `score` returned by the `/rerank` endpoint is _absolute_. This means that it is possible to
+compare `scores` across different calls to the `/rerank` endpoint. Each `score` is deterministic and represents how
+closely matched each specific item is to the query string.
+
```
ssmith-pc, line 1089: Nice, suggested rephrasing:

"The relevance `score` represents the absolute rank of a given query and passage pair. Normalized between [0, 1], the score represents how closely relevant a specific item and query are, with scores closer to 1 indicating high relevance."

ssmith-pc (approved):

README.md
aulorbe: Oo perfect, thank you!

ssmith-pc: It does represent that though, for example if I made 3 separate rerank calls with different data, but all the same query, and then I wanted to combine the results of all 3 calls and "order" them according to most relevant, then I could use the `score` field to determine their ranking position

aulorbe: Right, we could use `score` as a proxy for `rank`, but I don't think they're inherently equal, right? Like, a score represents the similarity between the document and the query string. A rank represents the returned document's order in the results list. Right now, they're one and the same, but what if the user wants to rerank (in the future) according to something other than relevance (e.g. time + relevance)? I think equating the two terms won't give us needed flexibility in the future. Thoughts?

ssmith-pc: How about:

> "The ~_relevance_~ score represents the absolute **measure of relevance** ~_rank_~ of a given query and passage pair. Normalized between [0, 1], the score represents how closely relevant a specific item and query are, with scores closer to 1 indicating high relevance."

aulorbe: Love it!

PR #294, opened 2024-10-07
Refactor + bug fix for assertWithRetries mechanism

aulorbe:
## Problem

We uncovered a bug with the current implementation of a test helper called `assertWithRetries` where it does not seem to be acting as expected, resulting in hanging behavior.

## Solution

1. Since this helper function is likely useful to our users more generally, I moved it into `utils` for easier discoverability.
2. I changed the signature of the function to just focus on retries (instead of retries + asserts). I think the function itself is cleaner this way, separates concerns, and is simpler to test. One can still pass whatever`assertions` they want to the `retry` function by way of the `asyncFn` param, which is still in the signature.
3. I changed the `<=` to `<`, since that seemed to be the root of the problem -- the function would never enter the final `else` statement, since when `attempts` === `maxRetries`, it became truthy and returned a resolved Promise, even though that Promise's value was `undefined`. 
4. I added unit tests to ensure this function works as expected.
5. A comment in the original code let me to believe the intent w/this function was to have an _exponential_ `delay` in it, but the current `delay` was actually multiplicative; it's now actually exponential.

## Type of Change

## Test Plan

New unit tests pass + existing integration tests that made use of `assertWithRetries` pass w/the new function.

---
  - https://app.asana.com/0/0/1208439494339500

### DIFF
--- a/src/utils/__tests__/retry.test.ts
+++ b/src/utils/__tests__/retry.test.ts
@@ -2,8 +2,8 @@ import { sleep } from '../../integration/testHelpers';
 import { retry } from '../retry';
 
 // Mock only `sleep`, keep other functions in testHelpers.ts as-is
-jest.mock('../testHelpers', () => ({
-  ...jest.requireActual('../testHelpers'),
+jest.mock('../../integration/testHelpers', () => ({
+  ...jest.requireActual('../../integration/testHelpers'),
   sleep: jest.fn().mockResolvedValue(undefined), // Directly mock sleep here
 }));
 
--- a/src/utils/__tests__/retry.test.ts
+++ b/src/utils/__tests__/retry.test.ts
@@ -1,0 +1,56 @@
+import { sleep } from '../../integration/testHelpers';
+import { retry } from '../retry';
+
+// Mock only `sleep`, keep other functions in testHelpers.ts as-is
+jest.mock('../testHelpers', () => ({
+  ...jest.requireActual('../testHelpers'),
+  sleep: jest.fn().mockResolvedValue(undefined), // Directly mock sleep here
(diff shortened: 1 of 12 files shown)


src/utils/__tests__/retry.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  ...jest.requireActual('../testHelpers'),
+  sleep: jest.fn().mockResolvedValue(undefined), // Directly mock sleep here
+}));
+
+describe('retry function', () => {
+  beforeEach(() => {
+    jest.clearAllMocks(); // Clear mocks between tests
+  });
+
+  // Unit test
```
aulorbe, line 15: Oops, I'll remove these

austin-denoble: Yeah I'd remove all the comments, I think?

src/utils/retry.ts
```diff
@@ -0,0 +1,24 @@
+import { sleep } from '../integration/testHelpers';
+
+export const retry = async (
```
austin-denoble, line 3: Does this actually need to move to the `/utils/` folder rather than remain in `testHelpers.ts`. This function was written explicitly for retrying flakey integration tests, so it feels like it makes more sense there at the moment.

This folder doesn't expose anything to users, nothing in here is exported out.

austin-denoble (commented):
Seems like integration tests have become much flakier with this change, or am I reading that wrong. The old `assertWithRetries` utility was primarily helpful in allowing time for freshness to update when trying to assert on things like upserts, deletes, etc. 

This isn't a utility we should be trying to expose to users.

PR #295, opened 2024-10-08
Change multiplicative delay to exponential delay

aulorbe:
## Problem

The comment in our current delay in `assertWithRetries` leads me to believe we want the delay to be exponential, but currently it's multiplicative.

## Solution

Change it to exponential :) 

## Type of Change

## Test Plan

pass.

---
  - https://app.asana.com/0/0/1208494629514994


aulorbe:
Exponential delay exceeds Jest's max timeout. This was just a side quest to make the code match the comment. Will leave as-is.

PR #299, opened 2024-10-11
Add ability for users to pass custom fields into /rerank

aulorbe:
## Problem

The current /rerank implementation does not allow users to rank on custom fields (they can only rank on the `'text'` field). We want to change that.

## Solution

Enable the following logic: 

**Route A: if user does NOT pass in a `rankFields` object:**
- If `documents` is a list of strings:
    - Set `rankFields` object to `['text']`
    - Transform list of strings to list of objs with `'text'` as key 
- If `documents` is a list of objects:
    - Set `rankFields` to the object's key(s)*

**Route B: if the user passes in a custom `rankFields` object:**
- If `documents` is a list of strings:
   - Set `rankFields` to user's custom field
   - Transform list of strings to list of objs where key == the user's custom field
- If `documents` is a list of objects:
  - Confirm each item in the `documents` contains the custom `rankField` field(s)

*Note: right now, the backend does _not_ error-out as expected when the user's docs have more than 1 key that is passed to the BGE model for reranking. [They are pushing a fix for this soon](https://pinecone-io.slack.com/archives/C06HPKVKUFM/p1728671996517259).

## Type of Change


src/inference/inference.ts
```diff
@@ -72,10 +71,10 @@ export class Inference {
...
    *  const myQuery = "What are some good Turkey dishes for Thanksgiving?";
    *  const myDocuments = [
-   *  { text: "I love turkey sandwiches with pastrami" },
-   *  { text: "A lemon brined Turkey with apple sausage stuffing is a classic Thanksgiving main" },
-   *  { text: "My favorite Thanksgiving dish is pumpkin pie" },
-   *  { text: "Turkey is a great source of protein" },
+   *  { myField: "I love turkey sandwiches with pastrami" },
+   *  { myField: "A lemon brined Turkey with apple sausage stuffing is a classic Thanksgiving main" },
+   *  { myField: "My favorite Thanksgiving dish is pumpkin pie" },
+   *  { myField: "Turkey is a great source of protein" },
```
gdj0nes, line 77: Should this still be `text` because it doesn't look like you are passing `rankFields`

jseldess:
> If documents is a list of objects:
Set rankFields to the object's key(s)*

@aulorbe, @gdj0nes, for route A, I think this isn't quite right. If the user does not pass `rankFields`, and does pass multiple objects, shouldn't we look for the object with `text` as the key? And if none have that key, error?

PR #303, opened 2024-10-15
Allow users to pass customField to /rerank endpoint

aulorbe:
## Problem

The current implementation of `/rerank` in the TS client does not (correctly) allow users to pass a custom field upon which to rerank. 

## Solution

Allow custom fields!

Please reference this PR to account for all expected functionality: https://github.com/pinecone-io/python-plugin-inference/pull/21/files

## Type of Change

## Test Plan

CI passes + reviewer xreferences PR above w/functionality introduced in this PR.

---
  - https://app.asana.com/0/0/1208523729730914


src/inference/inference.ts
```diff
@@ -140,42 +140,50 @@ export class Inference {
...
+          // If there is no `text` key:
+          throw new PineconeArgumentError(
+            'One or more documents are missing the specified rank field: `text`'
+          );
+        }
+      } else {
+        throw new PineconeArgumentError(
+          'Documents must be a list of strings or a list of objects'
+        );
+      }
```
austin-denoble, line 178: Is it possible for it to be a mix of the two? I'm actually not sure, but I'm assuming we don't allow that.

src/inference/inference.ts
```diff
@@ -140,42 +140,50 @@ export class Inference {
...
+      if (documents.every((doc) => typeof doc === 'string')) {
+        // Set rankFields to default value of 'text'
+        computedRankFields = ['text'];
+      } else if (
+        // If`documents` is an array of objects:
+        documents.every((doc) => typeof doc === 'object' && doc !== null)
+      ) {
+        // If there is a `text` key (other keys can be present too):
+        if (
+          documents.every((doc) => (doc as { [key: string]: string })['text'])
```
austin-denoble, line 165: Is there maybe a more efficient way to parse the `documents` that are passed and make some determinations about the contents one time up front rather than calling `documents.every` so many times? In the last code path we'd be iterating over everything in the array 3 times which is a bit heavy.

I'd try maybe looping over it once and building some parameters/bools that we use in these if statements rather than iterating repeatedly.

src/inference/inference.ts
```diff
@@ -140,42 +140,50 @@ export class Inference {
...
+
+    // If user does not pass in rankFields:
+    if (!rankFields) {
+      // If `documents` is an array of strings:
+      if (documents.every((doc) => typeof doc === 'string')) {
+        // Set rankFields to default value of 'text'
+        computedRankFields = ['text'];
+      } else if (
+        // If`documents` is an array of objects:
+        documents.every((doc) => typeof doc === 'object' && doc !== null)
```
austin-denoble, line 161: Why do we check `doc !== null` here but not with the `string` case above?

src/inference/inference.ts
```diff
@@ -140,42 +140,50 @@ export class Inference {
...
-            throw new PineconeArgumentError(
-              `The \`rankField\` value you passed ("${field}") is missing in the document at index ${index}`
-            );
-          }
-        });
-      });
-    }
+    // Standardize documents to ensure they are in object format
+    const newDocuments = documents.map((doc) =>
+      typeof doc === 'string' ? { text: doc } : doc
```
austin-denoble, line 185: Will the `text` key here be different if they passed a custom `rankFields`?

src/inference/__tests__/embed.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    const expected = [{ text: 'input1' }, { text: 'input2' }];
+    const result = inference._formatInputs(inputs);
+    expect(result).toEqual(expected);
+  });
+});
+
+describe('Inference Class: _formatParams', () => {
+  test('Should format params correctly', () => {
+    const params = { inputType: 'text', truncate: 'END' };
+    const expected = { inputType: 'text', truncate: 'END' };
```
austin-denoble, line 25: Do we need to test if we're not expecting this to change anything? Maybe I'm unclear on the test.

src/integration/inference/rerank.test.ts
```diff
@@ -34,4 +34,99 @@ describe('Integration Test: Pinecone Inference API rerank endpoint', () => {
...
+    const myDocuments = [{ text: 'doc1' }, { text: 'doc2' }];
+    expect(pinecone.inference.rerank(model, query, myDocuments)).toBeDefined();
+  });
+
+  test('Confirm docs as list of objects with additional customField + no rankfields, succeeds', async () => {
+    const myDocuments = [
+      { text: 'hi', customField: 'doc1' },
+      { text: 'bye', customField: 'doc2' },
+    ];
+    expect(pinecone.inference.rerank(model, query, myDocuments)).toBeDefined();
```
austin-denoble, line 63: I'm a little rusty on ts at this point, if the call fails it would avoid the expect and fail the test, right? Just trying to understand what this is asserting.

src/inference/__tests__/rerank.test.ts
```diff
@@ -0,0 +1,115 @@
...
+beforeAll(() => {
+  const config: PineconeConfiguration = { apiKey: 'test-api-key' };
+  const infApi = inferenceOperationsBuilder(config);
+  inference = new Inference(infApi);
+});
+
+test('Confirm throws error if no documents are passed', async () => {
+  const expectedError = new PineconeArgumentError(
+    'You must pass at least one document to rerank'
+  );
```
austin-denoble, line 20: nit: Personal preference, but I'd maybe just leave the expected error inline on the `expect().toEqual()`. Feels easier to parse.

austin-denoble (approved):
LGTM overall, I did leave some question and comments about a few things. I feel like I haven't read much TypeScript in a while and am a bit rusty. 😅

jseldess (commented):
Looks good to me, @aulorbe, but can you add an example to the README?

src/inference/inference.ts
aulorbe: Ah that's a good q! Right now, the TS client doesn't account for that. I'll add it in, thank you!

aulorbe:
> Looks good to me, @aulorbe, but can you add an example to the README?

Thank you!

I think [the existing example in the README on the RC branch](https://github.com/pinecone-io/pinecone-ts-client/blob/rc/2024-10/README.md) is still valid, right, or were you think I should add more to that one?

src/inference/inference.ts
ssmith-pc: Do you mean if someone mixed strings and documents in the same request? That would be kinda odd, I don't think we need to bend over backwards to support that

jseldess:
> > Looks good to me, @aulorbe, but can you add an example to the README?
> 
> Thank you!
> 
> I think [the existing example in the README on the RC branch](https://github.com/pinecone-io/pinecone-ts-client/blob/rc/2024-10/README.md) is still valid, right? Or were you think I should add more to that one?

Yeah, I think it's still valid. But perhaps we should add rankFields to the second example. Also, I'm not sure if it's worth NOT setting the other rank options in the first example (top_n, return documents, etc.). I'd suggest the first example has everything but rankfields, and the second example shows how to use rankfields to rank on a custom field.

src/inference/inference.ts
aulorbe: okay cool. I think my final `else` stmt would actually account for this anyways, no?

The logic is:
- if docs is an array of strings, do something
- else if docs is an array of objs, do something else
- else, error out

src/inference/inference.ts
aulorbe: I don't think that's a valid scenario, here's my thinking: 
- If someone passes in an array of strings, there are no keys in their documents
- If there are no keys in their documents, on the backend we artificially transform their documents into an array of key:value pairs, with a `text` key, simply b/c that's what the model needs for reranking
- If someone passes in an array of strings and then also passes in a custom field to rerank on... their docs would inherently be missing that custom field (by way of not being key:value pairs), so it would error out

Thoughts @ssmith-pc ?

src/inference/inference.ts
austin-denoble: Yeah, I think the logic should be fine for that atm.

src/integration/inference/rerank.test.ts
aulorbe: oh man, extremely good catch! I'll change to assert that rerankUnits are >= 1

src/inference/inference.ts
aulorbe: Added above!

src/inference/__tests__/rerank.test.ts
aulorbe: updated!

src/inference/__tests__/embed.test.ts
aulorbe: weird... this must be a holdover from something... I'll remove method (it does nothing currently) + this test. thanks!

aulorbe:
> > > Looks good to me, @aulorbe, but can you add an example to the README?
> > 
> > 
> > Thank you!
> > I think [the existing example in the README on the RC branch](https://github.com/pinecone-io/pinecone-ts-client/blob/rc/2024-10/README.md) is still valid, right? Or were you think I should add more to that one?
> 
> Yeah, I think it's still valid. But perhaps we should add rankFields to the second example. Also, I'm not sure if it's worth NOT setting the other rank options in the first example (top_n, return documents, etc.). I'd suggest the first example has everything but rankfields, and the second example shows how to use rankfields to rank on a custom field.

Cool, will do!

README.md
```diff
@@ -1131,15 +1131,40 @@ console.log(response);
...
+  },
+  {
+    title: 'Lemon Turkey',
+    body: 'A lemon brined Turkey with apple sausage stuffing is a classic Thanksgiving main',
+  },
+  {
+    title: 'Thanksgiving',
+    body: 'My favorite Thanksgiving dish is pumpkin pie',
+  },
+  { title: 'Protein Sources', body: 'Turkey is a great source of protein' },
```
jseldess, line 1148:
```suggestion
  { 
      title: 'Protein Sources', 
      body: 'Turkey is a great source of protein' 
  },
```

jseldess (approved):
README LGTM, with a nit

src/inference/inference.ts
```diff
@@ -140,41 +170,31 @@ export class Inference {
...
-        rankFields.forEach((field) => {
-          if (!(field in doc)) {
-            throw new PineconeArgumentError(
-              `The \`rankField\` value you passed ("${field}") is missing in the document at index ${index}`
-            );
-          }
-        });
-      });
+    if (!options.rankFields) {
+      // If each obj in newDocs does not have a 'text' field, throw an error
```
ssmith-pc, line 190: nit: the comment here should probably read "if _any_ obj does not have... " 

But may be more clear as simply:
"Ensure each obj in newDocs has a 'text' field"

src/inference/inference.ts
```diff
@@ -140,41 +170,31 @@ export class Inference {
...
+      if (
+        !newDocuments.every((doc) => typeof doc === 'object' && 'text' in doc)
+      ) {
```
ssmith-pc, line 193: The server will validate that the documents contain the fields they're supposed to, so it's not entirely necessary to validate on the client. However, if we want to validate on the client, why not also validate the case where a custom rank field was provided?

src/inference/inference.ts
```diff
@@ -140,41 +170,31 @@ export class Inference {
...
     }
-    // Destructure `options` with defaults
-    // Note: If the user passes in key:value pairs in `options` that are not the following, they are ignored
+
     const {
       topN = documents.length,
       returnDocuments = true,
-      rankFields = ['text'],
       parameters = {},
+      rankFields = undefined,
```
ssmith-pc, line 178: Do we need to destructure rankFields and set to undefined? Isn't it undefined by default? I think I don't entirely understand why we can't keep the `rankFields = ['text']` default

src/inference/inference.ts
ssmith-pc: I'm not fully following the original question since the impl has changed, but:

> * If there are no keys in their documents, on the backend we artificially transform their documents into an array of key:value pairs, with a text key, simply b/c

By "on the backend we transform" I think you mean, it's transformed in the client right? This is the right behavior

> * If someone passes in an array of strings and then also passes in a custom field to rerank on... their docs would inherently be missing that custom field (by way of not being key:value pairs), so it would error out

Correct, but it's fine if that error is detected in the server and sent back to the client. The client doesn't have to try and detect that.

src/inference/inference.ts
```diff
@@ -109,15 +141,13 @@ export class Inference {
...
-   * @param documents - (Required) The documents to rerank. Each document must be either a string or an object
-   * with (at minimum) a `text` key.
-   * @param options - (Optional) Additional options to send with the core reranking request. Options include: how many
-   * results to return, whether to return the documents in the response, alternative fields by which the model
-   * should to rank the documents, and additional model-specific parameters. See {@link RerankOptions} for more details.
+   * @param documents - (Required) An array of documents to rerank. The array can either be an array of strings or
+   * an array of objects.
+   * @param options - (Optional) Additional options to send with the reranking request. See {@link RerankOptions} for more details.
    * */
   @prerelease('2024-10')
```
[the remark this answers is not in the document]
aulorbe: Sure!

src/inference/inference.ts
aulorbe: oh yeah good call, I remove this part actually (the validation part)

src/inference/inference.ts
aulorbe: Ah good call, good call

austin-denoble (approved):
LGTM, thanks for iterating and helping me iterate in return! 🚢

ssmith-pc (approved):
Looks good!

src/inference/inference.ts
```diff
@@ -140,41 +168,24 @@ export class Inference {
           ' documents. See https://docs.pinecone.io/models for supported models.'
       );
     }
-    // Destructure `options` with defaults
-    // Note: If the user passes in key:value pairs in `options` that are not the following, they are ignored
+
     const {
       topN = documents.length,
       returnDocuments = true,
-      rankFields = ['text'],
```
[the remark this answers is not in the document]
aulorbe: (We spoke offline, I'm adding it back in simply to help users understand what's going on behind the scenes. TY @austin-denoble!)

PR #304, opened 2024-10-16
Add integration testing w/external NextJS app (Edge runtime)

aulorbe:
## Problem

We do not currently have a good way to test how our client behaves end to end in different environments. One of the chief problems our users have brought to our attention in the past is that some functionalities do not work in certain runtimes, e.g. Edge or Bun, and/or with certain frameworks, e.g. [NextJS](https://nextjs.org/docs).

## Solution

Build an external application in a runtime and framework known to have caused problems in the past and test our client from the end-user's perspective.

This PR introduces and end-to-end test suite that interacts with an external application written using the NextJS framework and the [Edge](https://vercel.com/docs/functions/edge-middleware/edge-runtime#edge-runtime) runtime.

Note: the app is automagically run in `Edge` by way of it using [middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware) (`middleware.ts`), and the [`Headers()` API](https://developer.mozilla.org/en-US/docs/Web/API/Headers).

## Overview of changes
- We now have [a public repo](https://github.com/pinecone-io/ts-client-e2e-tests) that contains a super simple application that builds a Pinecone serverless index, seeds it with data, and queries that data. We plan to add more operations in the near future; this is just a start. 
- This sample application creates an API endpoint that our client repo can send a POST request to and assert on the response. If the endpoint fails to deliver the expected response, we know that something is wrong with the most recent changes we are proposing to introduce in the client.
- This PR enables 2 types of interaction with the sample application:
    - Local end-to-end tests: for local runs, there is a new bash script that devs can execute. This will spin up the application on their `localhost:3000`.
    - CI end-to-end tests: for CI runs, we will now run this test automatically by way of adding it the `testing.yml` file that is run for each push to an open PR. In CI, the Github workflow and action hit the application's Vercel endpoint and assert on its response.
- There is a new CONTRIBUTING.md file with some info on this new test suite + the other existing ones, as well as a new README in the `external-app` dir itself with more in depth information.

## Type of Change

## Test Plan

If reviewers can pull down the code in this branch and try to run the tests locally, that'd be great.

## To Dos
I'll add a README and a CONTRIBUTING file to [the external app](https://github.com/pinecone-io/ts-client-e2e-tests)

---
  - https://app.asana.com/0/0/1208129226784956


.github/actions/e2e-testing/edge/action.yml
```diff
@@ -27,27 +31,47 @@ runs:
       uses: GuillaumeFalourd/clone-github-repo-action@main
       with:
         owner: 'pinecone-io'
-        repository: 'pinecone-rag-demo'
-        branch: 'main'
+        repository: 'ts-client-e2e-tests'
+        branch: 'Audrey/add-app'
```
aulorbe, line 35: I'll change this to `main` once I merge my PR in the app repo

.github/actions/e2e-testing/edge/action.yml
```diff
@@ -27,27 +31,68 @@ runs:
       uses: GuillaumeFalourd/clone-github-repo-action@main
```
aulorbe, line 31: Just a note for the previous step: `npm pack` packages up the existing code on the existing branch, so it'll run the code devs are proposing to merge in their PR, rather than whatever code is on `npm` etc. This removes the need for us to merge things before testing them end to end.

.github/actions/e2e-testing/edge/action.yml
```diff
@@ -27,27 +31,68 @@ runs:
...
+        indexName=$(echo "$response" | jq -r '.indexName')
+        echo "indexName=$indexName" >> $GITHUB_ENV
+
+        match_count=$(echo "$response" | jq '.queryResult.matches | length')
+        if [ $? -ne 0 ]; then
+          echo "Error: Request failed."
+          exit 1
+        fi
+
+        if [ "$match_count" -lt 1 ]; then
```
aulorbe, line 69: This is where we assert on the response in CI

.github/actions/e2e-testing/edge/action.yml
```diff
@@ -27,27 +31,68 @@ runs:
...
+    - name: Send POST request to endpoint Vercel app endpoint api/createSeedQuery
       run: |
-        cd pinecone-rag-demo
-        vercel --scope=pinecone-io --token=${{ inputs.vercel-token }} --yes
+        API_URL="https://ts-client-e2e-tests.vercel.app/api/createSeedQuery"
+        API_KEY="${{ inputs.PINECONE_API_KEY }}"
+
+        response=$(curl --silent --location --request POST "$API_URL" --header "PINECONE_API_KEY: $API_KEY")
+
+        indexName=$(echo "$response" | jq -r '.indexName')
```
aulorbe, line 60: Index names are randomized in the app. We then save the index name to the GITHUB_ENV environment variable file, so that we can delete them later -- this ensures we can have simultaneous runs of this workflow, from creating the index to deleting it later

src/end-to-end/edge-end-to-end.sh
```diff
@@ -0,0 +1,57 @@
...
+      echo "Error: Directory does not exist: $dir"
+      exit 1
+    fi
+elif [ "$clone" == "n" ]; then
+  echo "Not cloning ts-client-e2e-tests repo"
+fi
+
+# Compile and make a local link to the ts-client repo
+npm run build
+npm link
```
aulorbe, line 44: This will overwrite the portion in the app's `package.json` that says `"@pinecone-database/pinecone": "^3.0.4-rc.2024-10.0"`. Instead, it'll link to your local version.

src/end-to-end/edge-end-to-end.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Hop into ts-client-e2e-tests repo, install deps, link local ts-client repo, and start the Next.js server
+pushd "$dir/ts-client-e2e-tests"
+  npm install
+  npm link @pinecone-database/pinecone
```
aulorbe, line 49: This will overwrite the portion in the app's package.json that says "@pinecone-database/pinecone": "^3.0.4-rc.2024-10.0". Instead, it'll link to your local version.

You *must* run it after `npm install`, or else `npm install` will change `package.json` back to `"^3.0.4-rc.2024-10.0"`, since that's the latest version up on `npm` as of this writing.

src/end-to-end/edge-end-to-end.sh
```diff
@@ -0,0 +1,57 @@
...
+  npm install -g next@latest
+  next dev
+popd
+
+#npm uninstall -g next # Can comment out if do not want to uninstall the global Next.js package
```
aulorbe, line 54: We do have to install NextJS globally or else the `next dev` command (to spin up the app on your `localhost:3000` won't work when run from the `ts-client` repo, which is where this bash file is being executed. If you don't want Next globally, you can uncomment line 54 and uninstall it.

.github/actions/e2e-testing/edge/action.yml
austin-denoble: Would it be possible to port the logic for the actual testing to an explicit TypeScript / Javascript file that handles the calling / asserting, etc?

Something similar to `IntegrationRunner.js`. I think this will generally be easier to build out and manage if it lives in a code-based script rather than in a GH actions file.

CONTRIBUTING.md
```diff
@@ -0,0 +1,66 @@
...
+### Integration tests
+
+Simply run `npm run test:integration-local:<runtime>` to run all integration tests, substituting `<runtime>` with
+either `node` or `edge` as you see fit.
+
+### End-to-end tests
+
+Simply run `npm run test:end-to-end -- <clone y/n> <path>` to run all end-to-end tests, passing in `y` or `n` for
+whether you want to clone the repo that contains the external app used for testing, and the `path` the repo should be
+cloned in, or is already cloned in, e.g.:
```
austin-denoble, line 37: I'm not necessarily clear on why we there's an option to clone or not clone the test app repo. What's the difference here?

I would think we would just clone it every time to make sure we're running against whatever is fresh there.

src/end-to-end/edge-end-to-end.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+# Recommendation: Do not clone the ts-client-e2e-tests repo in local ts-client repo; it'll just cause npm deps issues
+
+# inputs
+clone=$1 # required -- `y` or `n` to clone the ts-client-e2e-tests repo
```
austin-denoble, line 8: This feels a little confusing considering the script has an option to clone or not clone. I asked a question about this previously but I feel like I'm missing something.

src/end-to-end/edge-end-to-end.sh
austin-denoble: I'd maybe add a comment to the script here stating the same thing.

.github/workflows/testing.yml
```diff
@@ -210,3 +210,17 @@ jobs:
         run: |
           cd semantic-search-example
           npm run test
+
+  end-to-end:
+    name: end-to-end
+    runs-on: ubuntu-latest
+    if: always() # Run teardown even if the tests fail
```
austin-denoble, line 217: nit: This isn't a teardown job though, right?

austin-denoble (commented):
Nice work getting this stood up, a lot of moving pieces here. I've left a few questions but we should maybe sync on this so you can walk me through some of the details. I feel like there's a couple things I'm confused about, but it's probably just lack of knowledge around some of this stuff.

CONTRIBUTING.md
aulorbe: Oh that's true, yes.... Okay I'll change!

.github/workflows/testing.yml
aulorbe: Ah must've been a c&p error, apologies! I'll remove.

.github/actions/e2e-testing/edge/action.yml
aulorbe: Definitely! Done. Now we have a file called `src/end-to-end/localRuns.ts`

.github/workflows/testing.yml
aulorbe: Ah, no, leaving this here, actually -- the e2e tests include a teardown job, so even if the test fails, the created index will be torn down, ensuring we don't just create orphan indexes in the event that these tests fail.

src/end-to-end/README.md
```diff
@@ -0,0 +1,33 @@
...
+    - The Vercel token is from the Pinecone Enterprise account and is stored in a Github secret
+- Action
+  - The action run by the workflow is called `actions/e2e-testing/edge/action.yml`
+  - This action packages the `ts-client` code on the current branch into the external app's repo
+  - It then installs the Vercel CLI and sends a POST request to the test app's `/api/createSeedQuery` endpoint,
+    which is up and running on Vercel
+  - If the response from the endpoint does not include a match (a match to a query composed by
+    the test app), the action fails
+  - The action then deletes the index that the test app spun up (note: it will attempt to delete the index even if
+    the assertions fail)
```
jhamon, line 26: This content doesn't seem that helpful to me. Since a lot of it is a narrative telling of the workflow, it seems likely to quickly go stale as CI configurations evolve. I would probably delete most of this, but add some explanatory comments in the workflow/job/action configs. Keeping that sort of documentation right next to the implementation will give us a better shot at keeping it up to date.

What I think would be useful here are:
- Some conceptual introduction to what these tests are and why we need them, how they differ from other integration tests, etc.
- Specific instructions for how to run locally. What environment variables are needed and how do I set them? 
- Links to the workflow source and github UI for the workflow.

README.md
```diff
@@ -1038,3 +1038,10 @@ generateQueryEmbeddings().then((embeddingsResponse) => {
 
 // << Send query to Pinecone to retrieve similar documents >>
 ```
+
+## Testing
+
+All unit, integration, and end-to-end testing takes place automatically in CI and is configured using Github actions
+and workflows, located in the `.github` directory of this repo.
+
+To run tests locally, see [CONTRIBUTING.md](CONTRIBUTING.md) for instructions.
```
jhamon, line 1047: I think this can probably be cut. Maybe leave a note directing people to CONTRIBUTING.md for the development readme. 

End-users trying to get their app built (the main audience for this readme) shouldn't ever need to run our tests. Also, explaining about the `.github` directory doesn't seem like the right level of granularity. Github Actions is in such widespread use on github that I think it feels like a waste of the reader's time to explain here. It's kind of like how we don't write a lot to explain other conventional aspects, like how we track module dependencies in our `package.json` file.

.github/actions/e2e-testing/edge/action.yml
```diff
@@ -27,27 +31,56 @@ runs:
...
+        indexName=$(ts-node src/end-to-end/localRuns.ts $API_KEY)
+             
+        if [[ "$indexName" != *"edge-test"* ]]; then
+          echo "Error encountered: $indexName"
+          exit 1
+        else
+          echo "Test succeeded! Index created: $indexName"
+        fi
+
+        echo "indexName=$indexName" >> $GITHUB_ENV
```
jhamon, line 69: This seems fine for now, but it would be cool to eventually have some sort of reusable action that let you do something like this:

```yaml
- name: Call Vercel test endpoints
  uses: ./actions/post-vercel-endpoint
  with:
    app_url: https://ts-client-e2e-tests.vercel.app
    vercel_token: '${{ secrets.vercel_token }}'
    endpoints: 
      - api/createSeedQuery
      - api/test2
      - api/test3
```

In particular it seems like the ability to parameterize on app_url would be useful, since right now we can only deploy and test one branch at a time.

jhamon (approved):
Seems like a solid start. There's a lot of moving pieces to this, so nice job.

.github/actions/e2e-testing/edge/action.yml
```diff
@@ -27,27 +31,56 @@ runs:
...
       shell: bash
 
-    - name: Deploy Project Artifacts to Vercel
+    - name: Clean up e2e indexes
```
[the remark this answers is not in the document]
aulorbe: I'll make a fast-follow ticket!

.github/actions/e2e-testing/edge/action.yml
aulorbe: cool!

austin-denoble (approved):
LGTM, thanks for taking on all the feedback and iterating on this. I'm really excited to have this test harness in place finally! 🚢

PR #313, opened 2024-11-01
Remove extra call to delete an already-deleted index in `bulkImport` integration test

aulorbe:
## Problem

We kept getting errors in CI that part of the `bulkImport` integration test was failing because it could not find the given index name. 

## Solution

Turns out, we accidentally left in an extra call to `delete` after we had already deleted the index! This PR just removes that call :)

## Type of Change

## Test Plan

CI passes w/flying colors 🚀 

---
  - https://app.asana.com/0/0/1208633891828391

### DIFF
--- a/src/integration/data/bulk/bulkImport.test.ts
+++ b/src/integration/data/bulk/bulkImport.test.ts
@@ -27,7 +27,6 @@ describe('bulk import', () => {
 
   afterAll(async () => {
     await retryDeletes(pinecone, indexName);
-    await pinecone.deleteIndex(indexName);
   });
 
   test('verify bulk import', async () => {


src/integration/data/bulk/bulkImport.test.ts
```diff
@@ -27,7 +27,6 @@ describe('bulk import', () => {
 
   afterAll(async () => {
     await retryDeletes(pinecone, indexName);
```
aulorbe, line 29: This function deletes the index, so it was already deleted once the following line executed, thereby throwing the `404` error

haruska (approved):

austin-denoble (approved):

PR #314, opened 2024-11-26
Small fix in deleteAll documentation regardning namespaces

kehanzhang:
## Problem

Describe the purpose of this change. What problem is being solved and why?
Was confusing when I read the commented documentation.

## Solution

[Describe the approach you took. Link to any relevant bugs, issues, docs, or other resources.](https://docs.pinecone.io/guides/data/delete-data#delete-all-records-from-a-namespace)
Change documentation to represent to default namespace being deleted.

## Type of Change

## Test Plan


aulorbe (approved):
Ah sorry for missing this, thank you so much for this great catch!

PR #318, opened 2024-12-19
Add retries to `configureIndex` and `update` operations

aulorbe:
## Problem

We first shipped retries to `upsert` as a POC. Now that we are happy with that, we are expanding retries to `configureIndex` and `update` operations.

This PR includes updates to the retry logic itself, too: I found some errors when I applied it to `configureIndex`, yay!

## Type of Change

## Test Plan

CI passes.

---
  - https://app.asana.com/0/0/1208718597634858

### DIFF
--- a/src/errors/http.ts
+++ b/src/errors/http.ts
@@ -198,6 +198,8 @@ export const mapHttpStatusError = (failedRequestInfo: FailedRequestInfo) => {
       return new PineconeInternalServerError(failedRequestInfo);
     case 501:
       return new PineconeNotImplementedError(failedRequestInfo);
+    case 503:
+      return new PineconeUnavailableError(failedRequestInfo);
     default:
       throw new PineconeUnmappedHttpError(failedRequestInfo);
   }
--- a/src/integration/control/configureIndex.test.ts
+++ b/src/integration/control/configureIndex.test.ts
@@ -1,16 +1,6 @@
-import {
-  BasePineconeError,
-  PineconeBadRequestError,
-  PineconeInternalServerError,
-} from '../../errors';
+import { BasePineconeError, PineconeBadRequestError } from '../../errors';
 import { Pinecone } from '../../index';
-import {
-  randomIndexName,
-  retryDeletes,
-  sleep,
-  waitUntilReady,
-} from '../test-helpers';
-import { RetryOnServerFailure } from '../../utils';
+import { randomIndexName, retryDeletes, waitUntilReady } from '../test-helpers';
 
 let podIndexName: string, serverlessIndexName: string, pinecone: Pinecone;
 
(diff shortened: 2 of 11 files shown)


src/errors/http.ts
```diff
@@ -198,6 +198,8 @@ export const mapHttpStatusError = (failedRequestInfo: FailedRequestInfo) => {
       return new PineconeInternalServerError(failedRequestInfo);
     case 501:
       return new PineconeNotImplementedError(failedRequestInfo);
+    case 503:
```
aulorbe, line 201: Whoops forgot to map this the 1st time around

src/integration/control/configureIndex.test.ts
```diff
@@ -75,26 +66,10 @@ describe('configure index', () => {
...
-        } catch (e) {
-          if (e instanceof PineconeInternalServerError) {
-            retryCount++;
-            await sleep(2000);
-          } else {
-            console.log('Unexpected error:', e);
-            throw e;
-          }
-        }
-      }
```
aulorbe, line 97: Don't need this now that we've got retries!

src/integration/data/vectors/upsertAndUpdate.test.ts
```diff
@@ -155,13 +157,47 @@ describe('Testing retry logic on Upsert operation, as run on a mock, in-memory h
     // Call Upsert operation
     await mockServerlessIndex.upsert(recordsToUpsert);
 
+    // 2 total tries: 1 initial call, 1 retry
+    expect(retrySpy).toHaveBeenCalledTimes(1); // passes
+    expect(delaySpy).toHaveBeenCalledTimes(1); // fails
+    expect(callCount).toBe(2);
+  });
+
+  test('Update operation should retry 1x if server responds 1x with error and 1x with success', async () => {
```
aulorbe, line 166: Figured duplicating this type of test across `upsert` and `update` would be enough to justify me _not_ duplicating it again for `configureIndex`, but lmk if you disagree!

aulorbe: (We should rly centralize this type of thing to avoid duplicating this logic, but I think this is okay for now)

src/integration/data/vectors/upsertAndUpdate.test.ts
```diff
@@ -96,13 +96,14 @@ describe('Testing retry logic on Upsert operation, as run on a mock, in-memory h
   let server: http.Server; // Note: server cannot be something like an express server due to conflicts w/edge runtime
   let mockServerlessIndex: Index;
   let callCount: number;
+  let op: string;
 
   // Helper function to start the server with a specific response pattern
   const startMockServer = (shouldSucceedOnSecondCall: boolean) => {
     // Create http server
-    server = http.createServer((req, res) => {
+    server = http.createServer({ keepAlive: false }, (req, res) => {
```
austin-denoble, line 104: Just curious, what happens when setting this to false?

src/utils/retries.ts
```diff
@@ -88,4 +106,24 @@ export class RetryOnServerFailure<T, A extends any[]> {
     // Ensure delay is not negative or greater than maxDelay
     return Math.min(maxDelay, Math.max(0, delay));
   };
+
+  private mapErrorIfNeeded(error: any): any {
+    if (error?.status) {
+      return mapHttpStatusError(error);
+    }
+    return error; // Return original error if no mapping is needed
```
austin-denoble, line 114: Do we know what errors we're seeing that end up without an associated `error.status`?

src/integration/data/vectors/upsertAndUpdate.test.ts
austin-denoble: It seems useful to validate that the calls themselves trigger retries as we'd expect, is that what you're talking about centralizing?

austin-denoble (approved):
Overall LGTM, it's nice that this is a pretty easy replacement due to how you set things up.

src/integration/data/vectors/upsertAndUpdate.test.ts
aulorbe: It simply ensures any outstanding http connections close once whatever you're doing on the spun up server concludes. It's sent as a default to `true` in Node20+, which is a new development. Unfortunately, setting it to `false` didn't fix the failing-tests-in-Node-20 problem :(

src/utils/retries.ts
aulorbe: I can't quite remember off the top of my head, and definitely need to look into this in the future at some point, but basically something about the BasePineconeError class sometimes has `status` in the json obj that's printed out in the console (if you print the error), but that shows up as `undefined` when you do `error.status`.

src/integration/data/vectors/upsertAndUpdate.test.ts
aulorbe: Mmm I'm not 100% sure we're on the same page -- when you say "the calls themselves," are you talking about the different async funcs that we could pass into the `RetryWrapper`?

aulorbe:
Re: CI/CD failures in Node20+:

I did a lot of research into why the `http` server we spin up in our integration tests fails in Node20+ but passes in Node18+. I honestly had to put a time-cap on my research into this, because I was finding _a lot_, but the TLDR is:
- In Node20+, the native `fetch` API stabilized. With that stabilization came a major upgrade to Node's internal `undici` dependency... from `5.x` to `7.x` in [this PR](https://github.com/nodejs/node/pull/56070).
- Within this upgrade were changes to how `undici` handles `http` connections wrt. You can see more [here](https://github.com/nodejs/node/pull/56070/files#diff-f516ab824a7722da938a4c7c851520d39731ddeb4f7198dff4e932c5d4f8fdf7R13128-R13152).

Basically, I had to add a `sleep` to our final integration test because I was getting the following error when I was debugging in Node20+, which indicates an aborted TCP connection:
```  
TypeError: fetch failed
        at node:internal/deps/undici/undici:13185:13 {
      [cause]: Error: read ECONNRESET
          at TCP.onStreamRead (node:internal/stream_base_commons:216:20) {
        errno: -54,
        code: 'ECONNRESET',
        syscall: 'read'
      }
    }
```

I tried a bunch of different ways to add a `sleep`, including adding timeouts to `server.listen` and `http.createServer`, as well as sleeps to the final `afterEach` method, but none fixed it. The only fix I could come up with was adding the `sleep` to the `Max retries exceeded w/o resolve` itself.

PR #319, opened 2024-12-23
Build out index tags tests a bit more

aulorbe:
## Problem

I just wanted to ensure that adding, deleting, and updating index tags on the same index gives us the results we expect, so added some integration tests.

## Test Plan

CI passes

### DIFF
--- a/src/integration/control/configureIndex.test.ts
+++ b/src/integration/control/configureIndex.test.ts
@@ -136,7 +136,7 @@ describe('configure index', () => {
       });
     });
 
-    test('Add/remove/update index tag(s) on serverless index', async () => {
+    test('Add/remove index tag(s) on serverless index', async () => {
       const description = await pinecone.describeIndex(serverlessIndexName);
       expect(description.tags).toEqual({
         project: 'pinecone-integration-tests',
--- a/src/integration/control/configureIndex.test.ts
+++ b/src/integration/control/configureIndex.test.ts
@@ -136,12 +136,13 @@ describe('configure index', () => {
       });
     });
 
-    test('Add index tag to a serverless index', async () => {
+    test('Add/remove/update index tag(s) on serverless index', async () => {
       const description = await pinecone.describeIndex(serverlessIndexName);
       expect(description.tags).toEqual({
         project: 'pinecone-integration-tests',
       });
 
+      // Add a tag
       await pinecone.configureIndex(serverlessIndexName, {
         tags: { testTag: 'testValue' },
       });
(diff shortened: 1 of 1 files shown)


src/integration/control/configureIndex.test.ts
```diff
@@ -136,17 +136,52 @@ describe('configure index', () => {
...
+        testTag: 'testValue',
       });
 
       await pinecone.configureIndex(serverlessIndexName, {
         tags: { project: '' },
       });
       const description2 = await pinecone.describeIndex(serverlessIndexName);
-      expect(description2.tags).toBeUndefined();
+      if (description2.tags != null) {
+        expect(description2.tags['project']).toBeUndefined();
```
iamannamai, line 167: This is interesting. I'm sure this has already been discussed, but I'd expect to have to explicitly pass `null` for this to happen and for empty strings to be the case, which is how a few other SDKs we use would do it.

src/integration/control/configureIndex.test.ts
```diff
@@ -136,17 +136,52 @@ describe('configure index', () => {
...
         project: 'pinecone-integration-tests',
+        testTag: 'testValue',
       });
 
       await pinecone.configureIndex(serverlessIndexName, {
         tags: { project: '' },
       });
       const description2 = await pinecone.describeIndex(serverlessIndexName);
-      expect(description2.tags).toBeUndefined();
+      if (description2.tags != null) {
```
iamannamai, line 166: If `undefined != null`, does this block still run? 

Separately, are tags always returned, even if they're empty? I guess that would be my expectation as a user of the SDK.

iamannamai (approved):
I don't really know enough about the expected behavior and that's really not the point of this PR, but love that the test just lets me asks these questions! If the behaviors are all as expected, good to go!

rajat08 (approved):

src/integration/control/configureIndex.test.ts
aulorbe: Question 1: `undefined` does equal `null`.

Question2 : Yes!

src/integration/control/configureIndex.test.ts
aulorbe: So the tag itself (the key value) is still present, but the value can be an empty string. If it's an empty string, then it's undefined. Does that make sense?

iamannamai: This is definitely more on the implementation side and agnostic to the SDK, but if I understand correctly that json responses don't really encode `undefined`, ie. the key is just omitted altogether, providing `project: ''` in this case would remove the key altogether? 

so i'd expect something like

```
input: `{ "project": null }`, output: `{}`
input: `{ "project": "" }`, output: `{ "project": "" }`
```

whereas this test tells me
```
input: `{ "project": "" }`, output: `{ }`
```

src/integration/control/configureIndex.test.ts
iamannamai: Ah ok, so the conflation of `null` and `undefined` are intentional here because tags should always be `{}` regardless of whether they've been added to the response or not? That would make sense to me!

iamannamai (commented):
ty for addressing the comments!

src/integration/control/configureIndex.test.ts
aulorbe: well, so `' '` removes the key, but the user can't pass `null` because of type safety in typescript (their IDE will tell them that the passed value in a tag must be a string)

aulorbe: (Same thing happpens here when you sub `undefined` for `null`:) 

Screenshot 2025-01-07 at 12 52 35 PM

PR #338, opened 2025-04-09
Allow `embed` as a parameter to `configureIndex`

austin-denoble:
## Problem
Currently, the `ConfigureIndexRequest` interface supports `embed?: ConfigureIndexRequestEmbed` as a parameter, which aligns with the OAS spec and how the endpoint can be used: https://github.com/pinecone-io/pinecone-ts-client/blob/e5b39d7a6c77836b5b0ae1f696ce7e6f6020f82e/src/pinecone-generated-ts-fetch/db_control/models/ConfigureIndexRequest.ts#L64

However, the `embed` field was not added to `ConfigureIndexRequestProperties`, defining what we expect a user to provide in a request. We also error out if the user attempts to configure an index with just the `embed` property.

## Solution
Update `ConfigureIndexRequestProperties` and the validation logic in the `configureIndex` action to allow leveraging `embed` properly.

## Type of Change

## Test Plan
CI Unit & Integration Tests


rohanshah18 (approved):
LGTM!

src/control/__tests__/configureIndex.validation.test.ts
```diff
@@ -37,7 +37,7 @@ describe('configureIndex argument validations', () => {
 
       await expect(toThrow).rejects.toThrowError(PineconeArgumentError);
       await expect(toThrow).rejects.toThrowError(
-        'Object contained invalid properties: speculoos. Valid properties include deletionProtection, spec, tags.'
+        'Object contained invalid properties: speculoos. Valid properties include deletionProtection, spec, tags, embed.'
```
jhamon, line 40: speculoos? lol

jhamon (approved):

src/control/__tests__/configureIndex.validation.test.ts
austin-denoble: Yeah, it's the invalid object key which ends up in the error message.

PR #397, opened 2026-07-14
ci: skipLibCheck in ts-compilation-test to fix TS 5.2–5.5 compat matrix

jhamon:
## Problem

The `TS compile` matrix (`.github/workflows/testing.yml`) is **red on every PR and on `main`** for the `~5.2.0`, `~5.3.0`, `~5.4.0`, `~5.5.0` legs. Example: [PR #395 checks](https://github.com/pinecone-io/pinecone-ts-client/pull/395/checks), [PR #394 checks](https://github.com/pinecone-io/pinecone-ts-client/pull/394/checks).

Root cause — **not** our published types. The compat harness does:

```yaml
npm install typescript@${{ matrix.tsVersion }} --no-cache
npm install @types/node --no-cache   # <-- unpinned → latest (26.x)
```

Recent `@types/node` `.d.ts` uses lib types that only exist in TS ≥ 5.6 (`IteratorObject`, `AsyncIteratorObject`, `BuiltinIteratorReturn`), so `tsc` fails **inside `node_modules/@types/node/*.d.ts`** on the 5.2–5.5 legs:

```
node_modules/@types/node/globals.d.ts(136,69): error TS2304: Cannot find name 'IteratorObject'.
node_modules/@types/node/globals.d.ts(141,74): error TS2552: Cannot find name 'AsyncIteratorObject'. ...
node_modules/@types/node/stream/web.d.ts(187,78): error TS2304: Cannot find name 'BuiltinIteratorReturn'.
```

The 5.6/5.7/5.8/latest legs pass — consistent with the diagnosis.

## Fix

Add `"skipLibCheck": true` to `ts-compilation-test/tsconfig.json`. This skips typechecking third-party `.d.ts` while **still fully checking the consumer's `src` against the built client's published types** — which is the entire point of this compat test. It does not weaken coverage of Pinecone's own types.

## Verification (local, no API key)

Reproduced with `typescript@~5.5.4` + `@types/node@26.1.1`:
- **Before:** 7 `error TS…` inside `node_modules/@types/node/*.d.ts` (identical to CI).
- **After `skipLibCheck: true`:** `0` `@types/node` errors.

Once merged, the 5.2–5.5 legs go green for #395, #394, and future PRs after they pick up `main`.

---

> [!NOTE]
> **Low Risk**
> Single CI harness tsconfig change with no runtime or published API impact; only affects how the compatibility test invokes `tsc`.
> 
> **Overview**
[...]

PR #395, opened 2026-07-13
test(smoke): mocked critical-path gate that runs keyless on every PR

jhamon:
## What

Propagates the Python SDK's key-free **mocked critical-path smoke gate** (PIN-75 §2) to the TypeScript SDK. This is the TS entry in the PIN-77 rollout (sequenced by traffic: TS → Go → Java → Rust).

## Why

The keyed integration suites can't run on forks / PRs without secrets, so a regression in the core request/response plumbing (host resolution, request building, response deserialization) can land unnoticed. This gate exercises the real **connect → upsert → query** path on every PR with **no API key**.

## How

- `src/smoke/mockedCriticalPath.test.ts` — drives `pc.index().upsert()` / `.query()` through the real client stack, with the HTTP transport mocked via `PineconeConfiguration.fetchApi` (the TS analogue of Python's respx). Asserts every leg of the critical path is actually hit, plus the host-cache behavior (one describe per index).
- `jest.config.smoke.js` + `npm run test:smoke:mocked` — dedicated runner targeting only `src/smoke/`. Does **not** set `--passWithNoTests`, so the job fails if the suite ever collects zero tests instead of silently passing.
- `.github/workflows/testing.yml` — new `Smoke (mocked, no key)` job that runs with no `PINECONE_API_KEY`.
- `jest.config.js` — excludes `src/smoke` from the unit run so it isn't double-collected.

## Verification

```
$ npm run test:smoke:mocked   # 2 passing, no API key in env
$ npx eslint src/smoke/ --ext .ts   # clean
```

Reference implementation: python-sdk `tests/smoke/test_mocked_critical_path_{sync,async}.py` + the `Smoke (mocked, no key)` job in `on-push.yml`.

Refs PIN-77.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Test and CI wiring only; no production SDK behavior changes beyond exercising existing `fetchApi` injection paths.
> 
> **Overview**
> Adds a **key-free CI smoke gate** that exercises the real client stack for **connect → upsert → query** on every PR, using a transport-level `fetchApi` mock (no `PINECONE_API_KEY`).
> 
> **`src/smoke/mockedCriticalPath.test.ts`** drives `Pinecone` → `index().upsert()` / `.query()` with routed mock responses for describe-index, upsert, and query; it asserts each HTTP leg is hit and adds a second case for **one describe per index** via `IndexHostSingleton` reset between tests.
> 
[...]

PR #394, opened 2026-07-11
security: npm update to resolve js-yaml medium Dependabot alert (#45)

jhamon:
## Summary
Resolves the single open medium Dependabot alert on pinecone-ts-client (PIN-28 / PIN-43).

| Alert | Package | Change | Sev |
|---|---|---|---|
| #45 | js-yaml | → 4.3.0 (transitive) | medium |

Applied via `npm update`. `npm audit` → 0 vulnerabilities after update.

## Verification
- `npm install` ✅  `npm audit` → 0 vulns ✅

---

> [!NOTE]
> **Low Risk**
> Dev-only dependency and lockfile changes with no runtime SDK logic changes; main risk is CI/tooling drift from the large transitive update.
> 
> **Overview**
> Refreshes **`package-lock.json`** via `npm update` to clear the medium **js-yaml** Dependabot alert (transitive dependency now resolves to **4.3.0** instead of 3.x).
> 
> In **`package.json`**, pins **`@types/node`** to exact **`22.19.7`** (was `^22.19.7`). A few **`src/`** edits are formatting-only (union types and `catch` blocks), with no behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30841d42211eac9eaced90151975e844864fd92b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #393, opened 2026-07-11
Enable Dependabot version updates (PIN-4)

jhamon:
## What
Adds `.github/dependabot.yml` enabling **weekly, grouped Dependabot version updates**.

Ecosystems covered: npm, github-actions.

## Why
Part of standing up the security fast-patch process (**PIN-4**). Keeping dependencies close to their latest patched release shrinks the diff and time needed to ship a fix when a security advisory lands. Dependabot **security updates** (auto patch PRs on advisories) and **vulnerability alerts** are already enabled on this repo; this adds the routine version-update cadence.

Minor/patch updates are grouped into a single PR per ecosystem to keep review noise low; majors come as individual PRs.

## Review notes
Config only — no code changes. First run may open a batch of update PRs; that is expected. Adjust `open-pull-requests-limit` or `schedule` if the cadence is too noisy.

---

> [!NOTE]
> **Low Risk**
> Config-only change with no runtime or auth impact; main effect is automated dependency PR volume.
> 
> **Overview**
> Adds **`.github/dependabot.yml`** so Dependabot runs **weekly version updates** for **npm** (repo root) and **github-actions**, alongside existing security alerts/advisory PRs.
> 
> npm **minor and patch** bumps are **grouped** into one PR (`npm-minor-and-patch`); **major** npm updates stay separate. GitHub Actions updates are grouped under `github-actions` (`patterns: '*'`). npm is capped at **10 open** Dependabot PRs at once.
> 
> No application code changes—CI/dependency maintenance config only. Expect an initial batch of update PRs on first run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef96f9a5257307b88c784b8614b516a18111ce5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #402, opened 2026-07-15
ci: skipLibCheck in ts-compilation-test to fix TS matrix compile

jhamon:
## Problem

The `testing / TS compile (~5.2.0 … ~5.5.0)` matrix jobs are red on **every** open PR — including config-only PRs like #393 (which touches only `.github/dependabot.yml`). That is the tell that the failure is not PR-specific but a shared breakage on `main`.

Root cause: the `install, compile, and run sample app` step in `.github/workflows/testing.yml` installs an **unpinned** `@types/node` (currently `26.x`). Its bundled lib definitions (`IteratorObject`, `BuiltinIteratorReturn`, updated `AsyncIterator` signatures) require **TS ≥ 5.6**, so `tsc` fails inside `node_modules/@types/node/*.d.ts` on the 5.2–5.5 matrix before the sample app's own code is ever checked.

Reproduced locally: `@types/node@26.1.1` + `typescript@5.2` → 8 errors in `@types/node` d.ts; adding `skipLibCheck` → exit 0.

## Fix

Enable `skipLibCheck` in `ts-compilation-test/tsconfig.json`. This skips type-checking of third-party `.d.ts` files while still fully type-checking the sample app against the client's published types — which is the actual purpose of this gate.

## Impact

Unblocks the TS-compile matrix on all currently-red PRs (#393, #394, #395, …). Rebase those onto `main` after this merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> CI-only TypeScript config tweak with no runtime or library behavior change.
> 
> **Overview**
> Adds **`skipLibCheck: true`** to `ts-compilation-test/tsconfig.json` so the CI **TypeScript 5.2–5.5 compile matrix** no longer fails on errors inside **`node_modules/@types/node`** (unpinned installs can pull typings that need TS ≥ 5.6).
> 
> The gate still type-checks the sample app under `src` against the published client types; only third-party declaration files are skipped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 068c49143c338cfe4b91038132bb3a6ad7946df2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ts-compilation-test/tsconfig.json
+++ b/ts-compilation-test/tsconfig.json
@@ -5,6 +5,7 @@
     "noUnusedLocals": true,
     "outDir": "lib",
     "rootDir": "src",
+    "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
     "target": "es2022",


jhamon:
Closing as a duplicate of #398, which is the canonical root-cause fix for this TS 5.2–5.5 compile-matrix breakage.

#398 pins `@types/node@20` (matching the job's Node 20 runtime), which keeps **full** lib type-checking — including the client's own published `.d.ts` — while resolving the version mismatch. `skipLibCheck` (this PR, and the earlier closed #397) would silence the errors but also stop type-checking declaration files, weakening the gate. #398 is the better fix and is already MERGEABLE/CLEAN.

